PR #3, opened 2022-01-11
[DEVHUB-948] Add consistent nav and footer.

mcolella14:
I copied over the theme a lot of styles and structured the layout like the current DevHub site. These will mostly change once we nail down the designs, but for now it at least looks decent with the existing styles.

https://user-images.githubusercontent.com/46664803/149027941-6fd56b52-03ef-4aa1-a821-a01e93e11323.mov


src/components/layout.tsx
```diff
@@ -0,0 +1,87 @@
...
+        background: ${colorMap.pageBackground};
+        color: ${colorMap.devWhite};
+        font-family: akzidenz, -apple-system, BlinkMacSystemFont, 'Segoe UI',
+            Helvetica, Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji',
+            'Segoe UI Symbol';
+        font-size: ${fontSize.default};
+        line-height: ${lineHeight.medium};
+        margin: 0;
+        padding: 0;
+    }
```
harikakotipalli, line 31: Not sure if it's emotion format, but can we have the styles in separate file ? I might be dumb here, but I always felt with this format , the main component and it's functionality are being lost with a lot of styling code above it.

mcolella14: Yeah I agree with you, it's a bit distracting when most of the file is styling. This is the way I've seen it in University/DevHub, but just did some research and there is also a convention of separating styled components into their own directory (for this is would be `src/components/layout.tsx` and `src/styled/layout.tsx`). Adds more files, but might make this clearer?

harikakotipalli: Correct. For me personally I prefer a separate file. @gfunk1230 any thoughts on this ?

gfunk1230: Yep! I would do that but I thought NextJS already had a built-in global file for CSS like `styles/globals.css` but if not I would create a global styles file and import that. Also, we should add CSS resets in that same file. Here is a good article on resets https://css-tricks.com/an-interview-with-elad-shechter-on-the-new-css-reset/

src/components/theme.ts
```diff
@@ -0,0 +1,270 @@
+import { css } from '@emotion/react';
+
+/**
+ * @type {Object.<string, string>}
+ * @property {string} fontsize returns px value
+ */
+const fontSize = {
```
gfunk1230, line 7: Having `src/components/theme.ts` is great but just something to keep in mind that this may all be replaced by the design system. For now, it's ok to keep it since we could use some parts of this.

gfunk1230 (approved):
Very minor comments but other than that LGTM!

package.json
```diff
@@ -9,11 +9,16 @@
     "format": "prettier --write ."
   },
   "dependencies": {
+    "@emotion/css": "^11.7.1",
```
gfunk1230, line 12: Very small nitpick but I tend to stay away from using the `~` or `^` so that in the case a  future minor/patch version introduces security vulnerabilities or causes and breakages. Not tied to that convention but something I have been wary of.

gfunk1230: Here is an article about it but I had a better before https://docs.renovatebot.com/dependency-pinning/

gfunk1230: Again not tied to this approach but was just was something practiced on University and previous projects

gfunk1230: This is another good article too https://erichurst.dev/when-to-pin-your-projects-dependencies/

src/components/layout.tsx
mcolella14: Yeah I think Next gets initialized with that global CSS file, but I just took it out since we were using emotion, and we rely on the theme for a lot of these styles, so it's nice to do the CSS in JS thing. I could just create a styled component in `src/styled/global.tsx` and then import that? And cool cool I'll look into resets!

src/components/theme.ts
mcolella14: Yeah I figured a lot of this will be changed out, just wanted something for now lol.

package.json
mcolella14: Ah ok that's interesting, pinning seems like a safe bet. Does Artifcatory kind of handle that though? I guess I'm not sure what the reason for using our own package registry is lol. I assumed the packages were like curated before being added to the registry, or is Artifactory solving some other problem?

harikakotipalli: @mcolella14 - as far as I remember we pointed to our internal registry only when consistent nav came into picture, which was developed by internal design systems team. All the other packages are downloaded from external.

PR #4, opened 2022-01-12
Devhub 949/cypress rt

harikakotipalli:
Adding cypress support


mcolella14 (approved):
Looks great! Pulled down and tested locally and everything is running smoothly. However, the ticket also included adding [React Testing Library](https://testing-library.com/docs/react-testing-library/intro/) to the project. We don't have any custom components at the moment though, so there's really nothing to test yet with React Testing Library. 

Just wondering if we should hold off until we have custom components to add React Testing Library, or add and configure it in this PR. @gfunk1230 what are your thoughts? 

Regardless, everything looks good here!

harikakotipalli:
Working on adding React testing library.

harikakotipalli:
Added React testing library setup.

mcolella14:
LGTM!

gfunk1230:
> Looks great! Pulled down and tested locally and everything is running smoothly. However, the ticket also included adding [React Testing Library](https://testing-library.com/docs/react-testing-library/intro/) to the project. We don't have any custom components at the moment though, so there's really nothing to test yet with React Testing Library.
> 
> Just wondering if we should hold off until we have custom components to add React Testing Library, or add and configure it in this PR. @gfunk1230 what are your thoughts?
> 
> Regardless, everything looks good here!

@mcolella14 sorry late to respond here but if we can get React testing library wired up that would be great.

gfunk1230:
@harikakotipalli Looks great!

gfunk1230 (approved):

PR #5, opened 2022-01-18
Devhub 951 subnavigation

harikakotipalli:
Adds initial skeleton for sub-navigation and placeholder for individual pages.


src/components/subnavigation/navbar.tsx
```diff
@@ -0,0 +1,44 @@
...
+    open: boolean;
+    setIsOpen: (open: boolean) => void;
+}
+
+export const Context = React.createContext<ContextProps>({
+    open: false,
+    setIsOpen: (open: boolean) => {
+        /**/
+    },
+});
```
mcolella14, line 18: Since it seems like we are only passing these values down 1 level (to `DeveloperTopic` and `DeveloperTopicsMenu`), what is the benefit of using Context here rather than just passing them as props?

src/components/articles.tsx
```diff
@@ -0,0 +1,16 @@
+import NavBar from './subnavigation/navbar';
+import React from 'react';
+import { PageWrapper } from '../styled/pages';
+
+const ArticlesComponent: React.FunctionComponent = () => {
+    return (
+        <>
+            <NavBar />
+            <PageWrapper>
```
mcolella14, line 9: If `NavBar` is going to be included in every page, I think it would make more sense to put it in `components/layout.tsx` below the consistent nav to keep more in line with Next.js conventions and avoid repetition.

src/components/subnavigation/developer-topics-menu.tsx
```diff
@@ -0,0 +1,29 @@
...
+        });
+    };
+
+    return (
+        <>
+            {open && (
+                <DeveloperTopicsMenuWrapper>
+                    <button onClick={handleDeveloperTopicsClick}>
+                        All Developer Topics
+                    </button>
```
mcolella14, line 22: Would we be able to just wrap the button in an anchor tag with an `href="/developertopics"` rather than use the next router manually? I'm not positive but I think it may be better for SEO to have anchor links when we can?

src/components/subnavigation/developer-topics.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+const DeveloperTopics: React.FunctionComponent = () => {
+    const { open, setIsOpen } = useContext(Context);
+    const handleOpening = () => {
+        setIsOpen(!open);
+    };
+    return (
+        <DeveloperTopicsTab>
+            <StyledButton onClick={handleOpening}>
+                Developer Topics
```
mcolella14, line 16: So I noticed on Figma that the developer topics menu also closes when we click outside of the menu. I'm not sure if that's the desired behavior in the real design, but just figured I'd point it out in case it comes up.

src/components/code-examples.tsx
```diff
@@ -0,0 +1,16 @@
...
+        <>
+            <NavBar />
+            <PageWrapper>
+                <h2>Code Examples</h2>
+            </PageWrapper>
+        </>
+    );
+};
+
+export default CodeExamplesComponent;
```
mcolella14, line 16: I think we could just put this all in `pages/articles.tsx` no? I don't think there's really a need to create a component for every entry in the `pages` directory, and it might be clearer to have the layout of the page in the actual `pages/articles.tsx` file rather than nested in another component.

mcolella14 (commented):
Only minor comments, everything looks good and I pulled and tested locally and it works great!

src/components/subnavigation/developer-topics-menu.tsx
harikakotipalli: Agree with you. Only used Router here to bring up discussion on it. Not sure what is the advantage of Next/Router over Links and a tags.

src/components/subnavigation/navbar.tsx
harikakotipalli: Correct. This goes back to your other comment "developer topics menu also closes when we click outside of the menu". If I have to handle that state change, can I still go with Prop drilling.

src/components/subnavigation/developer-topics-menu.tsx
mcolella14: Yeah I didn't know it was an option, I guess it could be cool for more complex routing in the future so good to know it's there.

src/components/subnavigation/navbar.tsx
mcolella14: Ah yeah that would sort of complicate it, I think maybe something could be done with [`useRef`](https://reactjs.org/docs/hooks-reference.html#useref) to help with this like is shown in [this post](https://stackoverflow.com/questions/32553158/detect-click-outside-react-component), but I think we can just cross that bridge when we come to it.

src/pages/articles.tsx
```diff
@@ -0,0 +1,7 @@
+import ArticlesComponent from '../components/articles';
+
+const Articles = (props: any) => {
```
gfunk1230, line 3: I assume the use of `any` for the `props` is temporary?

src/pages/codeexamples.tsx
```diff
@@ -0,0 +1,7 @@
+import CodeExamplesComponent from '../components/code-examples';
+
+const CodeExamples = (props: any) => {
```
gfunk1230, line 3: Same here I assume the use of `any` for the `props` is temporary?

src/styled/nav-bar.ts
```diff
@@ -0,0 +1,25 @@
+import styled from '@emotion/styled';
+import { colorMap } from './theme';
+
+const NavbarWrapper = styled('div')`
+    display: flex;
```
gfunk1230, line 5: Super nitpick but can we make sure to alphabetize the our CSS rules

src/styled/nav-button.ts
```diff
@@ -0,0 +1,10 @@
+import styled from '@emotion/styled';
+import { size } from './theme';
+
+const StyledNavList = styled('ul')`
+    padding: 0px ${size.xsmall} 0px ${size.xsmall};
```
gfunk1230, line 5: you can drop `px` when defining `0` as a value

src/styled/nav-bar.ts
```diff
@@ -0,0 +1,25 @@
...
+    flex-direction: row;
+    justify-content: space-around;
+    width: 100%;
+    position: relative;
+    transform: translateZ(0);
+    z-index: 2;
```
gfunk1230, line 11: We should consider creating constants for the levels of layers we want to support such as 

```
$zindex-dropdown:       1000;
$zindex-sticky:         1020;
$zindex-fixed:          1030;
$zindex-modal-backdrop: 1040;
$zindex-modal:          1050;
$zindex-popover:        1060;
$zindex-tooltip:        1070;
```

Article here https://www.smashingmagazine.com/2021/02/css-z-index-large-projects/

src/styled/nav-bar.ts
```diff
@@ -0,0 +1,25 @@
...
+    background: ${colorMap.greyDarkTwo};
+    div {
+        & > span {
+            font-size: medium;
+            font-family: Akzidenz-Grotesk Std;
```
gfunk1230, line 16: If we continue to reference this `font family `we should store it in a `const`.

src/styled/nav-bar.ts
```diff
@@ -0,0 +1,25 @@
+import styled from '@emotion/styled';
+import { colorMap } from './theme';
+
+const NavbarWrapper = styled('div')`
```
gfunk1230, line 4: I would consider using `nav` as the HTML tag versus a `div`

src/styled/developer-topics.ts
```diff
@@ -0,0 +1,16 @@
+import styled from '@emotion/styled';
+
+const StyledButton = styled('button')`
+    border: none;
+    background: none;
+    color: white;
```
gfunk1230, line 6: defining colors as `hex` values are faster than defining them as named colors. So `#fff `would be better

src/styled/developer-topics-menu.ts
```diff
@@ -0,0 +1,17 @@
...
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    width: 50%;
+    left: 25%;
+    background: ${colorMap.greyLightOne};
+    position: absolute;
+    z-index: 10;
+    height: 200px;
+    color: black;
```
gfunk1230, line 14: defining colors as `hex` values are faster than defining them as named colors. So `#000` would be better

src/styled/developer-topics-menu.ts
```diff
@@ -0,0 +1,17 @@
+import styled from '@emotion/styled';
+import { colorMap } from './theme';
+
+const DeveloperTopicsMenuWrapper = styled('div')`
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    width: 50%;
```
gfunk1230, line 8: Any reason we are using `percentage` values here versus leveraging `flex` to determine the amount of space to take?

src/styled/pages.ts
```diff
@@ -0,0 +1,13 @@
+import Styled from '@emotion/styled';
+import { colorMap } from './theme';
+
+const PageWrapper = Styled('div')`
+    background: white;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    height: 880px;
```
gfunk1230, line 9: Why are we setting a fixed height here?

src/styled/pages.ts
```diff
@@ -0,0 +1,13 @@
+import Styled from '@emotion/styled';
+import { colorMap } from './theme';
+
+const PageWrapper = Styled('div')`
+    background: white;
```
gfunk1230, line 5: Try not to use named colors but rather hex values. If we continue to use `white` and `black` often might be best to store `#000` and `#fff` in `const`.

src/components/articles.tsx
gfunk1230: Yep! good call @mcolella14

gfunk1230 (changes requested):
Looks good so far just a few request on changes

src/styled/nav-button.ts
```diff
@@ -0,0 +1,10 @@
+import styled from '@emotion/styled';
+import { size } from './theme';
+
+const StyledNavList = styled('ul')`
```
mcolella14, line 4: Should this be a `ul` tag? It looks like it just wraps the individual nav buttons, and it is causing lighthouse to fail because it expects only a `li` tag under `ul`.

src/components/subnavigation/hooks/use-outside-click.ts
```diff
@@ -0,0 +1,19 @@
...
+    useEffect(() => {
+        document.addEventListener('click', handleClick);
+
+        return () => {
+            document.removeEventListener('click', handleClick);
+        };
+    });
+};
+
+export default useOutsideClick;
```
mcolella14, line 19: Love this hook! But should we have this `hooks` directory at a higher level (`src/hooks`) for reusability throughout the project?

mcolella14 (approved):
Looks great! Just the Lighthouse thing to fix.

src/components/subnavigation/hooks/use-outside-click.ts
```diff
@@ -0,0 +1,19 @@
+import { useEffect } from 'react';
+
+const useOutsideClick = (ref: any, callback: any) => {
```
mcolella14, line 3: I think we should try to be explicit about these param types if it's possible.

gfunk1230 (approved):

PR #9, opened 2022-01-20
Add to readme and cleanup

mcolella14:
The main change here is adding `src/README.md` to explain the file structure. I also expanded the files that prettier affects, so the other changes are mostly formatting along with some cleaning up of imports.


gfunk1230 (approved):
Very Awesome!

PR #12, opened 2022-01-24
[DEVHUB-965] Add Apollo

mcolella14:
Added Apollo client and implemented it using apollo-rest-link to interact with the strapi dev api.

I just put the data in a list on the main page for now to demonstrate that it's there:
Screen Shot 2022-01-26 at 11 53 03 AM


src/requests/graphql/example.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const getCountries = async (): Promise<Country[]> => {
+    const { data } = await client.query({
+        query: gql`
+            query Countries {
+                countries(filter: { currency: { eq: "USD" } }) {
+                    name
+                    code
+                    emoji
+                }
```
harikakotipalli, line 14: You should be able to call rest API as well , without waiting for graphql layer to be ready. We are not sure if we want to add graphql or not at this point correct.

harikakotipalli: Documentation : https://www.apollographql.com/docs/react/api/link/apollo-link-rest/

mcolella14: Yeah we could hit a rest API, I kinda assumed that since we're implementing Apollo that we are going to go in the GraphQL direction, but I'm also open to just using rest endpoints if we want to write those on lambda functions.

harikakotipalli: Agreed - but I don't want us to wait for graphql layer to be ready. Also we donno if we get any benefit by adding Graphql to our backend. Why code when we don't have necessity :). We can start with strapi end points which we already have and then code new backend API's as we go in the best way we decide.

mcolella14: Ahh ok that makes sense to me. In the meantime, should we add that `apollo-link-rest`, or just fetch normally?

src/pages/index.tsx
```diff
@@ -1,23 +1,33 @@
...
 export default Home;
 
-// export async function getServerSideProps({}) {
-//     const allArticles = (await fetchApiRest('articles')) || [];
-//     return {
-//         props: { allArticles },
-//     };
-// }
+export const getServerSideProps: GetServerSideProps = async ({}) => {
+    const articlesModule = await import('../requests/articles'); // Dynamic import to allow unit testing without importing this.
```
mcolella14, line 28: Not exactly sure what the implications of this are on performance. I did this to keep jest from trying to instantiate the apollo client (I was having a lot of trouble mocking the module...), which it tried to do when the import was at the top of the file. I'm hoping the module will be loaded at build time rather than runtime this way, but not sure if that's the case.

If someone knows how to better mock it out with jest, that would be great.

src/requests/articles.ts
```diff
@@ -0,0 +1,19 @@
+import { gql } from '@apollo/client';
+
+import client from '../utils/apollo-client';
+import { Article } from '../interfaces/article';
+
+export const getArticles = async (): Promise<Article[]> => {
+    const { data } = await client.query({
+        query: gql`
+            query Articles {
+                articles @rest(type: "Article", path: "/articles") {
```
harikakotipalli, line 10: what is type here ? Response type ?

src/utils/apollo-client.ts
```diff
@@ -0,0 +1,11 @@
+import { ApolloClient, InMemoryCache } from '@apollo/client';
+import { RestLink } from 'apollo-link-rest';
+
+const restLink = new RestLink({ uri: process.env.STRAPI_URL });
+
+const client = new ApolloClient({
```
harikakotipalli, line 6: Can we call this strapiRestClient ?

harikakotipalli: Much better if we can create ClientFactory, which will give us the respective type of ApolloClient - Lambda Client, Strapi Client etc. Just a thought, this works for now but it we have more types, factory pattern works better.

src/pages/index.tsx
harikakotipalli: I din't understand the question, matt.

src/requests/articles.ts
mcolella14: Ah I will add that, thank you. Looks like it `ApolloQueryResult` https://www.apollographql.com/docs/react/api/core/ApolloClient/#ApolloClient.query

src/utils/apollo-client.ts
mcolella14: Ah I like the factory idea, I think that will resolve the issues that I was having with the tests as well.

.github/workflows/vercel-deploy.yml
```diff
@@ -21,5 +21,5 @@ jobs:
                   vercel-token: ${{secrets.VERCEL_TOKEN}}
                   vercel-org-id: ${{secrets.VERCEL_ORG_ID}}
                   vercel-project-id: ${{secrets.VERCEL_PROJECT_ID}}
-                  vercel-args: '--prod --build-env NPM_AUTH=${{ secrets.ARTIFACTORY_NPM_AUTH }} --build-env NPM_EMAIL=${{ secrets.ARTIFACTORY_NPM_EMAIL }}'
+                  vercel-args: '--prod --build-env NPM_AUTH=${{ secrets.ARTIFACTORY_NPM_AUTH }} --build-env NPM_EMAIL=${{ secrets.ARTIFACTORY_NPM_EMAIL }} --build-env STRAPI_URL=${{ secrets.STRAPI_DEV_URL }}'
```
harikakotipalli, line 24: this has to be changed to prod later.

src/utils/apollo-client.ts
```diff
@@ -0,0 +1,49 @@
...
+        });
+    }
+}
+class ApolloRestClient {
+    _client: ApolloClient<NormalizedCacheObject>;
+
+    constructor() {
+        this._client = new ApolloClient({
+            cache: new InMemoryCache(),
+            link: new RestLink({ uri: process.env.STRAPI_URL }),
```
harikakotipalli, line 28: what if you have to use another uri ? This should be passed, not coded inside.

src/utils/apollo-client.ts
```diff
@@ -0,0 +1,49 @@
...
+        });
+    }
+
+    getArticles = async (): Promise<Article[]> => {
```
harikakotipalli, line 32: getArticles should not be a client function.

src/utils/apollo-client.ts
harikakotipalli: to reuse for another URL

gfunk1230 (approved):

.github/workflows/vercel-deploy.yml
mcolella14: Yeah I figured we can change it when we actually put the data where it's supposed to go.

src/utils/apollo-client.ts
mcolella14: Where should it go? My thought was that it should be attached to the instance of the class in order to reuse the underlying apollo client, so we can utilize the cache and don't have to instantiate a new one every time we make a request.

harikakotipalli: it should be under a different folder api or data. getArticles primary job will be to construct the query, pass it to client instance. Client's job will be to run the query passed.

eg : getArticles = (client) => {
const query = ``;
const data = await client.query({query});
return data;
}

while testing you can 
mock client instance and mock await client.query call. 

about this part - we can utilize the cache and don't have to instantiate a new one every time we make a request - i think it is still valid in my case, I'm still using the same instance of the client. 

All said I'm okay with ever call you take.

harikakotipalli (approved):

src/utils/apollo-client.ts
mcolella14: Ok that makes sense to me. I've now separated out `getArticles` to it's own location (`src/requests/get-articles.ts`) where it accepts a client. Now when we get the static props, we just instantiate the client and call `getArticles` with the client. 

I've also added a unit test to this. It took me a while to get the mocking right, but it seems to be working now.

harikakotipalli (approved):
Awesome 👏 . Looks very good to me.

PR #22, opened 2022-01-27
DEVHUB-976 testing vercel deploy

harikakotipalli:
Adding vercel deployment workflow

### DIFF
--- a/.github/workflows/vercel-deploy.yml
+++ b/.github/workflows/vercel-deploy.yml
@@ -1,32 +1,12 @@
 name: Vercel Deploy
 on:
     push:
-    pull_request:
         branches: [main]
 env:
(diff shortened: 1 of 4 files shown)


.github/workflows/vercel-deploy.yml
```diff
@@ -0,0 +1,39 @@
+name: Vercel Deploy
+on:
+    pull_request:
+        branches: [main]
```
mcolella14, line 4: I think we'd want this to happen on a merge into main, not a PR, right?

mcolella14 (changes requested):
Awesome that this works! Just the `on: pull_request` thing, I think lighthouse is failing because `lighthouserc.js` is still configured to look in the `build` directory.

mcolella14 (approved):

PR #27, opened 2022-02-04
[DEVHUB-978] Hit lambda function instead of strapi directly.

mcolella14:
I will update the lambda functions to interact with the database directly, but that requires some setup in AWS, so the build is hitting the lambda function which is hitting strapi. [This is the lambda that it's hitting now](https://us-east-2.console.aws.amazon.com/lambda/home?region=us-east-2#/functions/devcenter-backend-ArticlesFunction-MGMb7MhgHMkV?tab=code). Also, to run locally will require an `env.local` file like this:
```
API_GATEWAY_URL=https://ohayd84u58.execute-api.us-east-2.amazonaws.com/Prod
```
Technically the Strapi URL isn't needed at the moment, but the API Gateway URL will be necessary to build.


gfunk1230 (approved):
LGTM!

PR #30, opened 2022-02-18
Devhub 1002/shared card component

harikakotipalli:


mcolella14:
The Lighthouse checks are failing because of [this heading tag check](https://web.dev/heading-order/). We are kinda tied to the tags because of the designs so I think we can ignore this. If you add this `'heading-order': 'warn',` to `lighthouserc.js` in the `assertions` object, it should just throw a warning.

mcolella14 (approved):
Looks great on the staging site! No blocking issues on my end, just suggestions and the lighthouse issue.

PR #32, opened 2022-02-18
Devhub 1003/search and sort components

mcolella14:
Jira Tickets: [Search/Sort](https://jira.mongodb.org/browse/DEVHUB-1003) and [Load More](https://jira.mongodb.org/browse/DEVHUB-1004)

@harikakotipalli I wanted to get your opinion on this file structure layout I did for this component:
Screen Shot 2022-02-18 at 5 05 02 PM
I saw that the flora repo follows this pattern of having tests/styles/types/utilities in the sam directory of the component, making it easier to keep track of what belongs to the component. I could switch back to our usual structure, but wanted to bring this up before we build out too many things.


src/components/search/utils.ts
```diff
@@ -0,0 +1,26 @@
+import { Fetcher } from 'swr';
```
harikakotipalli, line 1: Can we move this file to requests ?

harikakotipalli (approved):
Just one comment. LGTM!!.

src/components/search/utils.ts
mcolella14: So I think the actual fetch logic we will move to `requests` when we implement it, so like on line 19 we would just call a function imported from `requests`. But since this `Fetcher` type is specific to the `swr` hook, and since we're only using that in the search component, I think I'd still want the definition of `fetcher` to remain here.

harikakotipalli: okay, sure.

src/components/search/search.tsx
```diff
@@ -0,0 +1,90 @@
...
+            <TypographyScale variant="heading5" sx={titleStyles}>
+                All {name} Content
+            </TypographyScale>
+            <div sx={searchBoxStyles}>
+                <TextInput
+                    name="search-box"
+                    label={`Search ${name} Content`}
+                    iconName={ESystemIconNames.SEARCH}
+                    value={search}
+                    onChange={val => {
```
gfunk1230, line 60: nitpick but you can just swap `val` with either `e` or `event` since you already have to specifying that you're getting a value with the `value` property off of the `target` property. Hope this makes sense.

gfunk1230: In fact not a biggie but for performance and not creating a new function object on every typing of character the user does you can create a function like this:

```
const onSearch = (e) => {
        setSize(1);
        setSearch(val.target.value);
}
```

Then on the onChange prop just set it like so:

`onChange={onSearch}`

This way on every character update you're referencing the named function `onSearch` in memory versus creating a new function object. Hope this makes sense.

src/components/search/styles.ts
```diff
@@ -0,0 +1,34 @@
...
+    marginBottom: ['inc40', null, 'inc70'],
+};
+
+export const dataStyles = {
+    gridColumn: ['span 6', null, 'span 8', 'span 12', '4 / span 9'],
+    margin: 'auto',
+    display: 'flex',
+    flexDirection: 'column' as 'column', // theme-ui is weird about this.
+    alignItems: 'center',
+    width: '100%',
```
gfunk1230, line 26: Curious why we need to set the width at 100% if this is a block element and we are also determining the `width` already with `gridColumn`?

src/pages/topics/[slug].tsx
```diff
@@ -1,7 +1,12 @@
 import type { NextPage, GetStaticProps, GetStaticPaths } from 'next';
 import { ParsedUrlQuery } from 'querystring';
 
+import styled from '@emotion/styled';
+import theme from '@mdb/flora/theme';
+import { GridLayout } from '@mdb/flora';
+
 import Hero from '../../components/hero';
+import Search from '../../components/search/search';
```
Comment (gfunk1230) [inline src/pages/topics/[slug].tsx:9]: btw what do you think about at some point we create alias for common file directories so that we are not repeatedly writing out .`./../../etc `?

Comment (gfunk1230) [reply src/pages/topics/[slug].tsx:9]: Like so: https://webpack.js.org/configuration/resolve/

src/components/search/search.tsx
```diff
@@ -0,0 +1,90 @@
...
+            shouldRetryOnError: false,
+        }
+    );
+
+    return (
+        <>
+            <TypographyScale variant="heading5" sx={titleStyles}>
+                All {name} Content
+            </TypographyScale>
+            <div sx={searchBoxStyles}>
```
gfunk1230, line 54: You could leverage the form tag with a role of search so that we follow some accessibility best practices like illustrated here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search#search_form_labels_and_accessibility

gfunk1230 (approved):
@mcolella14 this looks great! I had some comments but approving

src/components/search/search.tsx
mcolella14: Ah ok that makes sense, but does that still apply if we are re-rendering the component whenever we type? I'll still make the change to be on the safe side, but just curious how react handles function creation on render.

src/pages/topics/[slug].tsx
Comment (mcolella14) [reply src/pages/topics/[slug].tsx:9]: Yeah i think we should absolutely do that, importing components from the pages is getting cumbersome lol.

src/components/search/styles.ts
mcolella14: Ah I think I left that in there by mistake, thanks for the catch!

PR #33, opened 2022-02-21
Devhub 1000/featured jumbo card

harikakotipalli:


src/components/cards/featured-card.tsx
```diff
@@ -0,0 +1,94 @@
+import React from 'react';
+import { Thumbnail } from '../../interfaces/thumbnail';
+import { PillCategory } from '../../types/pill-category';
+import { TypographyScale } from '@mdb/flora';
```
gfunk1230, line 4: nitpick but can we have third party libraries like `@mdb/flora` at the top just like we do with React

src/components/cards/featured-card.tsx
```diff
@@ -0,0 +1,94 @@
...
+featured jumbo card consumes a thumbnail, pill, header, paragraph, tags, and footer with the date and Appleseed section
+We will need to follow up with Michael Waltzer about the Appleseed section
+Tags would need to be supported via a prop that can map over the number of tags with its text
+The cards for Search filtered content we can reuse the code from the jumbo featured card built here. The only difference here is that this card extends 100 percent of the available space.
+Talked to Michael Waltzer about podcast cards and looks like we would like to keep them as is in the design.
+We can provide support to have a variant for podcast cards that only displays the thumbnail, pill, header, and footer with the date
+Authors section is conditional so will render if there are authors
+ */
+
+const myLoader = ({ src, width, quality }: any) => {
```
gfunk1230, line 40: can we typeset these props so that we are not using `any` ?

src/components/cards/featured-card.tsx
```diff
@@ -0,0 +1,94 @@
...
+};
+
+const FeatureCard: React.FunctionComponent<IProps> = ({
+    authors,
+    contentDate,
+    description,
+    title,
+    pillCategory,
+    tags,
+    thumbnail,
```
gfunk1230, line 51: since we are only referencing the `url` property from the `thumbnail` any chance we can just explicitly reference the thumbnail URL by having a prop named `thumbnailUrl` which receives the string of the `img` src

src/components/cards/featured-card.tsx
```diff
@@ -0,0 +1,94 @@
...
+                        variant="identifier"
+                        text={pillCategory}
+                        size="small"
+                    />
+                    <StyledTitle variant="heading6">{title}</StyledTitle>
+                    <StyledDescription variant="body2">
+                        {description}
+                    </StyledDescription>
+                    <TagWrapper>
+                        {tags?.map(tag => (
```
gfunk1230, line 76: Wouldnt we always receive a length from tags ? so if it's always expected would we then need `?` `optional chaining ` ?

src/components/cards/shared-card.tsx
```diff
@@ -0,0 +1,84 @@
...
+const myLoader = ({ src, width, quality }: any) => {
+    return `${src}?w=${width}&q=${quality || 75}`;
+};
+
+const SharedCard: React.FunctionComponent<IProps> = ({
+    contentDate,
+    description,
+    title,
+    pillCategory,
+    thumbnail,
```
gfunk1230, line 42: since we are only referencing the `url` and `size` property from the `thumbnail` any chance we can just explicitly reference the thumbnail `url` and `size`  by having a prop named `thumbnailUrl`  and thumbnailSize or we could destructure it?

src/pages/topics/[slug].tsx
```diff
@@ -25,6 +36,45 @@ const Topic: NextPage<TopicProps> = ({ name, description, ctas }) => {
...
                 ctas={ctas}
             />
+            <div
+                style={{
+                    display: 'flex',
+                    gap: '20px',
+                    flexWrap: 'wrap',
+                }}
+            >
+                {contentCategories.map(c =>
```
Comment (gfunk1230) [inline src/pages/topics/[slug].tsx:46]: nitpick but can we be more explicit when naming the arguments here so rather than just `c` we could have a `category`?

src/pages/topics/[slug].tsx
```diff
@@ -25,6 +36,45 @@ const Topic: NextPage<TopicProps> = ({ name, description, ctas }) => {
                 description={description}
                 ctas={ctas}
             />
+            <div
+                style={{
```
Comment (gfunk1230) [inline src/pages/topics/[slug].tsx:40]: I think using emotion's `css` prop might be better here versus style? 

https://emotion.sh/docs/object-styles

src/pages/topics/[slug].tsx
```diff
@@ -25,6 +36,45 @@ const Topic: NextPage<TopicProps> = ({ name, description, ctas }) => {
...
+                    display: 'flex',
+                    gap: '20px',
+                    flexWrap: 'wrap',
+                }}
+            >
+                {contentCategories.map(c =>
+                    getL1Content()
+                        .filter(a => a.pillCategory == c)
+                        .map(d => (
```
Comment (gfunk1230) [inline src/pages/topics/[slug].tsx:49]: Same here about being more explicit with the naming of arguments/parameters `d` in fact we can even use destructuring here like so `map(d => ({title, pillCategory, thumbnail, description, contentDate, title})`

src/pages/topics/[slug].tsx
```diff
@@ -25,6 +36,45 @@ const Topic: NextPage<TopicProps> = ({ name, description, ctas }) => {
...
+            <div
+                style={{
+                    display: 'flex',
+                    gap: '20px',
+                    flexWrap: 'wrap',
+                }}
+            >
+                {contentCategories.map(c =>
+                    getL1Content()
+                        .filter(a => a.pillCategory == c)
```
Comment (gfunk1230) [inline src/pages/topics/[slug].tsx:48]: Same here about making arguments more explicit since just declaring `a` isn't very descriptive

src/requests/get-l1-content.ts
```diff
@@ -0,0 +1,152 @@
+import { CardContent } from '../interfaces/cardContent';
+
+const getL1Content = (): CardContent[] => {
+    return [
+        {
+            pillCategory: 'DEMO APP',
+            thumbnail: {
+                size: 'medium',
+                url: 'https://mongodb-devhub-cms.s3.us-west-1.amazonaws.com/ATF_720x720_17fd9d891f.png',
```
gfunk1230, line 9: if your going to reference `https://mongodb-devhub-cms.s3.us-west-1.amazonaws.com/` in a few places I would just store it in a const like 

`const AWS_URL = 'https://mongodb-devhub-cms.s3.us-west-1.amazonaws.com/'`

src/styled/feature-card.ts
```diff
@@ -0,0 +1,129 @@
...
+import { PillCategory } from '../types/pill-category';
+import { HorizontalRule, Pill, Tag, TypographyScale } from '@mdb/flora';
+import Image from 'next/image';
+import { pillColorMap } from './pill-color';
+
+const FeaturedCardWrapper = styled('div')`
+    padding: 32px;
+    height: max-content;
+    border: 1px solid #e7eeec;
+    box-shadow: 0px 2px 4px rgba(0, 0, 0, 0.1);
```
gfunk1230, line 12: In CSS when declaring `0` as a value you don't need to add `px`

src/styled/feature-card.ts
```diff
@@ -0,0 +1,129 @@
...
+        height: 180px;
+        width: 180px;
+    }
+`;
+
+const StyledThumbnail = styled(Image)`
+    border-radius: 8px;
+`;
+
+const CardHeader = styled('div')`
```
gfunk1230, line 51: If this is meant to act as a header I would use the `header` HTML tag. Also what do when I am not sure what tag to use I go to HTML5 doctor website 

http://html5doctor.com/element-index/#header

src/styled/feature-card.ts
```diff
@@ -0,0 +1,129 @@
...
+    display: grid;
+    grid-template-columns: max-content 1fr;
+    grid-gap: ${theme.space.elementXSmall};
+    // Tablet
+    @media only screen and (min-width: ${theme.sizes.breakpoint.medium}) {
+        grid-gap: ${theme.space.elementMedium};
+    }
+`;
+
+const ContentWrapper = styled('div')``;
```
gfunk1230, line 61: I assume this was left by accident ?

src/styled/feature-card.ts
```diff
@@ -0,0 +1,129 @@
...
+const StyledHorizontalRule = styled(HorizontalRule)`
+    // Mobile
+    margin-top: ${theme.space.elementSmall};
+    // Desktop small
+    @media only screen and (min-width: ${theme.sizes.breakpoint.large}) {
+        margin-top: ${theme.space.elementMedium};
+    }
+`;
+
+const FooterContent = styled('div')`
```
gfunk1230, line 107: You could possibly leverage the `footer` HTML tag here?  http://html5doctor.com/the-footer-element-update/

```
<section>
   Section content appears here.
   <footer>
   Footer information for section.
   </footer>
</section>
```

src/styled/pill-color.ts
```diff
@@ -0,0 +1,7 @@
+export const pillColorMap = {
+    VIDEO: '#F9EBFF',
```
gfunk1230, line 2: Shouldn't we be using `tokens` here from `flora`?

src/styled/shared-card.ts
```diff
@@ -0,0 +1,90 @@
...
+import { HorizontalRule, Pill, TypographyScale } from '@mdb/flora';
+import { PillCategory } from '../types/pill-category';
+import { pillColorMap } from './pill-color';
+import Image from 'next/image';
+
+const SharedCardWrapper = styled('div')`
+    padding: 32px;
+    height: max-content;
+    border: 1px solid #e7eeec;
+    box-shadow: 0px 2px 4px rgba(0, 0, 0, 0.1);
```
gfunk1230, line 12: You can remove the `px` from the 0 declaration

src/styled/shared-card.ts
```diff
@@ -0,0 +1,90 @@
...
+    // Mobile
+    width: 467px;
+    // Desktop small
+    @media only screen and (min-width: ${theme.sizes.breakpoint.large}) {
+        width: 336px;
+    }
+`;
+
+const ThumbnailImage = styled(Image)`
+    border-radius: 8px;
```
gfunk1230, line 23: You have hard-coded `px` values in a couple of places for `padding`, `shadows`, `margins`, etc what I would do is look in `flora` if they use `tokens` for any of these values so we don't have to hard code them.

src/styled/shared-card.ts
```diff
@@ -0,0 +1,90 @@
...
+    margin-bottom: ${theme.space.elementXSmall};
+    // Desktop small
+    @media only screen and (min-width: ${theme.sizes.breakpoint.large}) {
+        margin-bottom: ${theme.space.elementSmall};
+    }
+`;
+
+const IntrinsicRatioWrapper = styled('div')`
+    position: relative;
+    width: 100%;
```
gfunk1230, line 37: This is a block element so you don't need to declare it to take 100% of the width.

src/styled/shared-card.ts
```diff
@@ -0,0 +1,90 @@
...
+type PillProps = {
+    pillCategory: PillCategory;
+};
+
+const StyledPill = styled(Pill)`
+    margin-bottom: ${theme.space.elementXSmall};
+    background-color: ${(props: PillProps) => pillColorMap[props.pillCategory]};
+    // Desktop small
+    @media only screen and (min-width: ${theme.sizes.breakpoint.large}) {
+        padding: 8px 24px;
```
gfunk1230, line 60: Same here about looking at the `tokens` in `flora` for some of these values

gfunk1230 (approved):
Left a few comments but approved so as not to block you. I would have @mcolella14 take a look as well. Looks good so far @harikakotipalli

gfunk1230:
I actually forgot another thing here. So I and @mcolella14 talked about the file structure of a component and thought it would make sense to have everything that belongs specifically to the component in the same file such as styles and tests. These is something back in my day lol which was considered the ducks approach which you can read about it here 

https://medium.com/building-crowdriff/react-redux-file-architecture-ducks-it-up-6b32eaaba341#:~:text=The%20basic%20idea%20behind%20Ducks,that%20is%20several%20levels%20deep.

src/components/cards/featured-card.tsx
mcolella14: @gfunk1230 tags is an optional prop, so I believe this will just short circuit if tags is undefined.

src/styled/feature-card.ts
```diff
@@ -0,0 +1,129 @@
...
+import theme from '@mdb/flora/theme';
+import { PillCategory } from '../types/pill-category';
+import { HorizontalRule, Pill, Tag, TypographyScale } from '@mdb/flora';
+import Image from 'next/image';
+import { pillColorMap } from './pill-color';
+
+const FeaturedCardWrapper = styled('div')`
+    padding: 32px;
+    height: max-content;
+    border: 1px solid #e7eeec;
```
mcolella14, line 11: We should probably use the flora theme value for this color (`colors.black20 = '#e7eeec'`).

src/styled/feature-card.ts
mcolella14: `Represents the "header" of a document or section of a document.` I don't really think these cards are really functioning as a section of the document.

src/styled/feature-card.ts
mcolella14: Same my comment about the headers.

mcolella14 (approved):
Looks good to me! Nothing blocking from my end, just some questions.

PR #34, opened 2022-02-24
Test lighthouse explicit urls

mcolella14:


harikakotipalli (approved):
@mcolella14 @gfunk1230 
Agreeing this part will be manual, picking the content heavy pages for running light house checks against to get worst case metrics.

PR #37, opened 2022-02-25
Clean up file structure with the Hero component, fix link animations.

mcolella14:
Migrating from using styled components to `sx`, also refactored to the new file structure for components


harikakotipalli (approved):

src/components/card-section/card-section.tsx
```diff
@@ -16,22 +20,22 @@ const CardSection: React.FunctionComponent<CardSectionProps> = ({
...
+            right: [
+                `calc(${theme.sizes.inc60} - ${theme.sizes.inc70})`,
+                null,
+                `calc(${theme.sizes.inc50} - ${theme.sizes.inc70})`,
+            ],
+        });
+
+    const onLinkLeave = () => setHoverStyles({});
+
+    // Memoize so we don't re-render the cards when this hover stuff happens.
```
gfunk1230, line 36: Awesome! I like this

src/components/card-section/card-section.tsx
```diff
@@ -44,10 +48,54 @@ const CardSection: React.FunctionComponent<CardSectionProps> = ({
...
+        >
+            <div sx={sectionHeadingTopStyles}>
+                <TypographyScale variant="heading5">{title}</TypographyScale>
+                <div
+                    sx={linkWrapperStyles}
+                    onMouseEnter={onLinkEnter}
+                    onMouseLeave={onLinkLeave}
+                >
+                    <Link
+                        href="#"
```
gfunk1230, line 68: I believe we can remove `href` altogether to make this render as a button versus a traditional link.

gfunk1230: Also I should be able to add `onMouseEnter={onLinkEnter} onMouseLeave={onLinkLeave}` ?

src/components/hero/styles.ts
```diff
@@ -0,0 +1,48 @@
...
+};
+
+export const breadcrumbStyles = {
+    alignItems: 'center',
+    display: 'flex',
+    gap: 'elementXSmall',
+};
+
+export const linkStyles = {
+    'div > span.textlink-default-text-class': {
```
gfunk1230, line 18: I would be careful with targeting the HTML tag here since if the tag changes these CSS rules wouldn't apply anymore. If there is already a class on it I would target that alone.

gfunk1230 (approved):
Looks great! Just made some suggestions. Other than that good to merge.

src/components/card-section/card-section.tsx
mcolella14: Rendering as a button would have SEO impacts though right? 

And coincidentally, the `onMouseEnter` and `onMouseLeave` props are [only used in `Link` if it's rendering as a button](https://github.com/10gen/flora/blob/main/src/Link/Link.tsx#L104), so removing the href would solve having to have the div wrap this.

And yeah if we did that I think we could add it on the link itself, but would have to test it out.

PR #38, opened 2022-03-04
Populate L1 Page with cards

mcolella14:
[JIRA Ticket](https://jira.mongodb.org/browse/DEVHUB-1012)

We now have a single Card component that accepts a variety of variants, each with different behavior based on viewport size. We don't have these cards linking to anything yet, but it shouldn't be hard to implement.


harikakotipalli (approved):
Looks great to me.

PR #39, opened 2022-03-09
Devhub 1010/tertiary nav

mcolella14:
[Jira Ticket](https://jira.mongodb.org/browse/DEVHUB-1010)

Gonna ask mike about the scrolling behavior on the small desktop nav, I just made some assumptions on scroll behavior and did a POC here but will see what he says and change it accordingly.

I also bumped flora, and there were some breaking changes (things that had `box-sizing: border-box` seem to not have that attribute anymore?), but those should be fixed in this.

PR #40, opened 2022-03-10
L1/L2 variants

mcolella14:
[Jira Ticket](https://jira.mongodb.org/browse/DEVHUB-1016)

A lot of dummy data here, but main thing was making changes based on the content variants.

Heavy Variant : `/developer/topics/atlas`

Medium Variant: `/developer/topics/data-lake`

Small Variant: `/developer/topics/vs-code`


src/pages/topics/[slug].tsx
```diff
@@ -49,32 +57,82 @@ const Topic: NextPage<TopicProps> = ({
...
+                                    )}
+                                </>
+                            )}
+                        </GridLayout>
+                    </div>
+                )}
+                <Search name={name} slug={slug} />
+
+                {variant === 'light' && relatedTopics.length > 0 && (
+                    <div>
```
Comment (gfunk1230) [inline src/pages/topics/[slug].tsx:126]: do we need this `div` here?

src/pages/topics/[slug].tsx
```diff
@@ -49,32 +57,82 @@ const Topic: NextPage<TopicProps> = ({
...
+                                    topics={topics}
+                                    title={`${name} Topics`}
+                                />
+                            )}
+                            {variant !== 'light' && (
+                                <FeaturedCardSection content={featured} />
+                            )}
+                            {variant === 'heavy' && (
+                                <>
+                                    {tutorials.length > 2 && (
```
Comment (gfunk1230) [inline src/pages/topics/[slug].tsx:87]: Let me know what you think about making this into a map so that we don't have to repeat the same code. If you think its an overkill I am ok with leaving it as is. 

```
{[tutorials, demoApps, articles, videos, podcasts,].map((section,i) => {
    if (section.length > 2) {
        return (
            <CardSection
                key={section.category}
                content={section}
                title={section.category}
            />
        )
    }
}
```

gfunk1230 (approved):
Left very minor comments but LGTM!

src/pages/topics/[slug].tsx
Comment (mcolella14) [reply src/pages/topics/[slug].tsx:87]: Ah so i actually converted it into a map in [this PR](https://github.com/mongodb/devcenter/pull/39). I'll merge main into this branch and that should resolve it.

src/pages/topics/[slug].tsx
Comment (mcolella14) [reply src/pages/topics/[slug].tsx:126]: I was having some issues with `GridLayout` having some weird margins, and we wrapped the upper topics container grid in a div as well, so I just did that here. But I'll see if I can sort it out because yeah having excessive elements is annoying.

PR #47, opened 2022-03-29
[DEVHUB-1014] Content page header with feedback flow.

mcolella14:
[Jira Ticket](https://jira.mongodb.org/browse/DEVHUB-1014)

The bulk of the work here is actually the rating feedback flow. It's a bit hard to fully implement without the backend in place, but I think I've done as much as I can in the meantime.

I used [Greg's draft PR](https://github.com/mongodb/devcenter/pull/42) as a starting point, so some of the changes are from that.

Also, I created a `pages/[...slug].tsx` file. This is the catch-all route, and will map all the slugs in the products to routes, so content will be available at it's slug. Ex. for an article, go to `/product/atlas/a1` and for a video go to `/product/atlas/v1`.


src/components/feedback-modal/dialogs/thank-you.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+const ThankYou: React.FunctionComponent<{ onContinue: () => void }> = ({
+    onContinue,
+}) => {
+    const title = 'We appreciate your feedback.';
+    const subtitle = (
+        <>
+            We&apos;d love to chat with you and answer your questions in our
+            online{' '}
+            <a
```
gfunk1230, line 13: Just wanted to double-check we don't want to use the Link component for this?

src/pages/[...slug].tsx
```diff
@@ -0,0 +1,352 @@
...
+                        }}
+                        layout="fill"
+                    />
+                </div>
+            )}
+            {!vidOrPod && ratingSection}
+        </>
+    );
+
+    const contentFooter = (
```
Comment (gfunk1230) [inline src/pages/[...slug].tsx:151]: Did we mean to keep this here?

gfunk1230 (approved):
LGTM! just minor comments

src/components/feedback-modal/dialogs/thank-you.tsx
mcolella14: Ah yeah so I actually ran into the same issue as Harika with the Links not being inline, so I just did this, but I'll try to fix it, thanks for catching.

src/pages/[...slug].tsx
Comment (mcolella14) [reply src/pages/[...slug].tsx:151]: Yeah it's a work in progress since the footer will also go on this page.

src/components/feedback-modal/dialogs/thank-you.tsx
mcolella14: Actually just gonna merge and fix this in my next PR.

PR #49, opened 2022-04-01
Devhub 1015/article content footer

mcolella14:
[Jira Ticket](https://jira.mongodb.org/browse/DEVHUB-1015)

The main components here are the `SeriesCard` and `RequestContentModal`.  I did make some minor changes to other areas:

- Changed the `href` prop on the `Card` component to `slug` to be more explicit about what to render, because the behavior of `next/link` gets confused by our routing.
- Made some styling changes to the `FeedbackModal` to keep it in line with what I did in `RequestContentModal`.

### DIFF
--- a/src/components/request-content-modal/dialogs/text-request.tsx
+++ b/src/components/request-content-modal/dialogs/text-request.tsx
@@ -4,12 +4,16 @@ import { TextRequestProps } from '../types';
 import { modalWrapperStyles } from '../styles';
 import { useState } from 'react';
 
+const helperText =
+    "Please include which products, languages, and technologies you're interested in seeing as part of this content request";
+
 const TextRequest: React.FunctionComponent<TextRequestProps> = ({
     onContinue,
     contentCategory,
 }) => {
-    const helperText =
-        "Please include which products, languages, and technologies you're interested in seeing as part of this content request";
+    const title = `Request ${
+        /^[aeiou]/gi.test(contentCategory) ? 'an' : 'a'
+    } ${contentCategory}`; // Regex to tell if it starts with a vowel.
 
     const [topic, setTopic] = useState('');
     const [description, setDescription] = useState('');
@@ -24,8 +28,7 @@ const TextRequest: React.FunctionComponent<TextRequestProps> = ({
                     marginBottom: ['inc40', null, null, 'inc50'],
                 }}
             >
-                Request {/^[aeiouh]/gi.test(contentCategory) ? 'an' : 'a'}{' '}
-                {contentCategory}
+                {title}
             </TypographyScale>
 
             <div
--- a/src/components/request-content-modal/dialogs/text-request.tsx
+++ b/src/components/request-content-modal/dialogs/text-request.tsx
@@ -1,0 +1,84 @@
+import { TypographyScale, Button, TextArea, TextInput } from '@mdb/flora';
+
+import { TextRequestProps } from '../types';
+import { modalWrapperStyles } from '../styles';
+import { useState } from 'react';
+
+const TextRequest: React.FunctionComponent<TextRequestProps> = ({
+    onContinue,
+    contentCategory,
+}) => {
+    const helperText =
+        "Please include which products, languages, and technologies you're interested in seeing as part of this content request";
+
+    const [topic, setTopic] = useState('');
+    const [description, setDescription] = useState('');
+    const [email, setEmail] = useState('');
+
+    return (
+        <div sx={modalWrapperStyles}>
+            <TypographyScale
+                variant="heading5"
+                sx={{
+                    display: 'block',
+                    marginBottom: ['inc40', null, null, 'inc50'],
+                }}
+            >
+                Request {/^[aeiouh]/gi.test(contentCategory) ? 'an' : 'a'}{' '}
+                {contentCategory}
+            </TypographyScale>
+
+            <div
+                sx={{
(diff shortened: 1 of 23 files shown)


src/pages/[...slug].tsx
```diff
@@ -149,139 +164,52 @@ const ContentPage: NextPage<ContentPiece> = ({
...
-                        >
-                            Request a _____
-                        </Button> */}
-        </>
+                    ))}
+                </Grid>
+            </div>
+            <div sx={{ display: 'flex', justifyContent: 'center' }}>
+                <Button
+                    onClick={() => setRequestContentModalStage('text')}
```
Comment (gfunk1230) [inline src/pages/[...slug].tsx:206]: Assuming this is acting as a placeholder?

src/components/request-content-modal/request-content.modal.tsx
```diff
@@ -0,0 +1,44 @@
...
+            <form sx={{ width: '100%' }}>
+                {modalStage === 'text' ? (
+                    <TextRequest
+                        onContinue={(topic, description, email) => {
+                            setModalStage('thanks');
+                            onSubmit(topic, description, email);
+                        }}
+                        contentCategory={contentCategory}
+                    />
+                ) : modalStage === 'thanks' ? (
```
gfunk1230, line 34: Couldn't we do this rather? 

```
modalStage === 'thanks' && (
                    <ThankYou onContinue={() => setModalStage('closed')} />
                ) 
```

gfunk1230: Or like this too ->

```
<form sx={{ width: '100%' }}>
                {modalStage === 'text' ? (
                    <TextRequest
                        onContinue={(topic, description, email) => {
                            setModalStage('thanks');
                            onSubmit(topic, description, email);
                        }}
                        contentCategory={contentCategory}
                    />
                ) : modalStage === 'thanks' ? (
                    <ThankYou onContinue={() => setModalStage('closed')} />
                ) : null}
            </form>
```

gfunk1230 (approved):

src/components/request-content-modal/dialogs/text-request.tsx
```diff
@@ -0,0 +1,87 @@
...
+                    marginBottom: ['inc40', null, null, 'inc50'],
+                }}
+            >
+                {title}
+            </TypographyScale>
+
+            <div
+                sx={{
+                    '& > div': {
+                        width: '100%',
```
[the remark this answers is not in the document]
mcolella14: This is targeting the `TextInput` component, which normally has a fixed width, but we want it to take up the whole width.

src/components/request-content-modal/dialogs/text-request.tsx
```diff
@@ -0,0 +1,87 @@
...
+                    onChange={e => setTopic(e.target.value)}
+                />
+            </div>
+
+            <div sx={{ marginBottom: 'inc30' }}>
+                <TextArea
+                    name="description"
+                    label="Describe your experience"
+                    value={description}
+                    onChange={e => setDescription(e.target.value)}
```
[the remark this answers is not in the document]
mcolella14: So I've been reading about the performance of callbacks like this, and [this article](https://dmitripavlutin.com/dont-overuse-react-usecallback/) kinda made me think in terms of "if it becomes a problem we'll fix it", rather than overcomplicating the code. Basically we'd have to write a `useCallback` hook for each field, and I'm not super convinced that avoiding defining a handful of functions is worth the complexity.

I just profiled it and it really doesn't look like much (<1ms).

src/components/request-content-modal/request-content.modal.tsx
```diff
@@ -0,0 +1,44 @@
...
+const RequestContentModal: React.FunctionComponent<
+    RequestContentModalProps
+> = ({ modalStage, setModalStage, contentCategory }) => {
+    const onSubmit = (topic: string, description: string, email: string) => {
+        const contentRequest: ContentRequest = {
+            topic,
+            description,
+            email,
+        };
+        console.log(contentRequest);
```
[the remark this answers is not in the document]
mcolella14: Right now we don't have a backend in place to post these requests, so I was planning on leaving this here just as some tangible way of seeing the post body until we have something in place.

src/pages/[...slug].tsx
Comment (mcolella14) [reply src/pages/[...slug].tsx:206]: No, `text` is a modal stage representing the stage where we prompt the user for text input.

src/components/series-card/series-card.tsx
```diff
@@ -0,0 +1,78 @@
...
+    const { content, title } = series;
+    let currentPieceIndex;
+    for (let i = 0; i < content.length; i++) {
+        if (content[i].slug === currentSlug) {
+            currentPieceIndex = i;
+            break;
+        }
+    }
+    if (currentPieceIndex === undefined) {
+        return <></>; // The slug should be in the series.
```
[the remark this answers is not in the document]
mcolella14: So `currentPieceIndex` should never be undefined, but we have to handle the case where it is for typescript to be happy. I wasn't really sure how to fail here, so I just went with the soft fail. But we could also just throw an error if this is the case. Which do you think would be better?

src/components/request-content-modal/request-content.modal.tsx
mcolella14: I will do this, forgot you can return null for JSX.

PR #52, opened 2022-04-08
[DEVHUB-1046] Topic Content Type Pages

mcolella14:
I will probably refactor this after Harika's changes are merged, since there are some changes to the same files going on.

Main differences:

- I've created a `page-templates` folder, which contains the templates for both the Topic Landing pages and the Topic Content Type pages. This is necessary because they share the same page file (`[l1_l1]/[...slug]`). Now, `getStaticPaths` will create the content type pages for every topic route we create. For example, when we create the route for `/product/atlas`, we check all of the atlas content, see what content types it has, and create routes for all of those (`/product/atlas/articles`, `/product/atlas/videos`, etc.). Then, in `getStaticProps`, we check if this route is a content type page or a landing page, pass that to the page, and return the correct template.
- Also made changes to the topics cards and side nav to actually get the correct routing in there. This may change depending on Harika's implementations.
- Made the Search component more flexible and added tests to cover.


src/page-templates/topic-landing/topic-landing.tsx
```diff
@@ -0,0 +1,171 @@
...
+                                'span 6',
+                                null,
+                                'span 8',
+                                'span 12',
+                                '4 / span 9',
+                            ],
+                        }}
+                    />
+
+                    {variant === 'light' && relatedTopics.length > 0 && (
```
gfunk1230, line 159: Just a recommendation to shorthand this but still fine as is 
`{variant === 'light' && !!relatedTopics.length && (`

gfunk1230 (approved):
Looks awesome @mcolella14! Just made suggestions but LGTM otherwise

PR #58, opened 2022-04-22
Devhub 1044

harikakotipalli:


package.json
```diff
@@ -28,19 +28,20 @@
...
         "remark": "^13.0.0",
         "remark-directive": "^1.0.1",
         "remark-gfm": "^1.0.0",
         "sanitize-html": "^2.7.0",
         "styled-components": "5.3.3",
         "swr": "1.2.1",
         "theme-ui": "0.10.0",
-        "unist-util-visit": "^2.0.3",
-        "react-player": "^2.9.0"
+        "unist-util-visit": "^2.0.3"
```
mcolella14, line 44: Should probably pin these just to be safe.

src/interfaces/author.ts
```diff
@@ -0,0 +1,9 @@
+type Image = {
+    url: string;
+};
+
+export interface Author {
+    name: string;
+    bio: string;
+    image: Image;
+}
```
mcolella14, line 9: Won't we need a `url` or `slug` field for the authors in order to link to their corresponding author page?

src/interfaces/tag-type-response.ts
```diff
@@ -0,0 +1,25 @@
...
+
+interface GenericTagTypeResponse {
+    name: string;
+    calculatedSlug: string;
+}
+
+export interface ExpertiseLevelTag extends GenericTagTypeResponse {}
+
+export interface ContentTypeTag {
+    contentType: PillCategory;
```
mcolella14, line 11: Can we just call this field `name` rather than `contentType` to keep it in line with the structure of the generic tag interface?

src/types/pill-category.ts
```diff
@@ -3,4 +3,6 @@ export type PillCategory =
     | 'Demo App'
     | 'Tutorial'
     | 'Podcast'
-    | 'Video';
+    | 'Video'
+    | 'Quickstart'
+    | 'Code Example';
```
mcolella14, line 8: Not necessary, but should we think about renaming this from `PillCategory` to something more informative? Like `ContentType`? I think it could get confusing for someone reading this in the future.

mcolella14 (approved):
Overall looks great! No blocking comments, just nits.

mcolella14:
Looks like lighthouse is having an issue with having a link to the "Go" language page. It thinks that "Go" isn't descriptive because that usually just means go somewhere, so I think we'll have to add `link-text` to the fields we ignore in the lighthouse config.

src/interfaces/tag-type-response.ts
harikakotipalli: the way I created strapi , I cudn't manipulate that field on graphql, thats why this had to be an exception

src/interfaces/author.ts
harikakotipalli: Good catch.will add.

PR #60, opened 2022-04-26
DEVHUB-1063 backend data for secondary navigation

harikakotipalli:


src/service/get-secondary-nav-menu.ts
```diff
@@ -0,0 +1,138 @@
...
+        quickstartMenu,
+        codeExamplesMenu,
+        podcastsMenu,
+        videosMenu,
+    ];
+};
+
+const getL2ForL1 = (l2Items: Tag[], l1Name: string) => {
+    return l2Items.filter(l2Item =>
+        l2Item.slug.toLowerCase().split('/').includes(l1Name.toLowerCase())
```
mcolella14, line 91: This logic could be an issue if we were to have an L1 with a space in the name. I think we should search be the L1 slug instead right? Ex. if we had an L1 product with the name `Cloud Manager`, the slug would probably be `cloud-manager`, but here we'd include the space by just doing `toLowerCase` on the name: `Cloud Manager` -> `cloud manager`. If we just pass the slug in as the second param to this function, i think it would would correctly in that case.

mcolella14 (commented):
Looks great! Just the one edge case.

src/service/get-secondary-nav-menu.ts
harikakotipalli: Great catch. !! you are right, I should be checking for slug.

mcolella14 (approved):

gfunk1230 (approved):

PR #61, opened 2022-04-27
Content Type pages with filters

mcolella14:
The filters here are the bulk of the work. The parsing of the filters looks intensive, but it's actually fast after getting all the content.

I also implemented the real search functionality on both the existing search component and new pages. 

New pages will be at `/articles`, `/tutorials`, etc.

This is kinda a huge PR, but mainly a lot of refactoring in addition to the new stuff.

@gfunk1230 The card to use on the homepage is `components/showcase-card`


src/components/search-filters/desktop/filters.tsx
```diff
@@ -0,0 +1,57 @@
...
+            )}
+            {!!technologyItems.length && (
+                <FilterGroup
+                    title="Technology"
+                    items={technologyItems}
+                    filters={allFilters}
+                    setFilters={onFilter}
+                />
+            )}
+            {/* <FilterGroup
```
gfunk1230, line 39: I assume we can delete this?

gfunk1230 (approved):
Looks great! I just left one coment

src/components/search-filters/desktop/filters.tsx
mcolella14: Yeah but it may be implemented in the future so I think we can leave it for now.

PR #64, opened 2022-05-02
Devhub 1044 - Topic Content Type pages Refactor

harikakotipalli:


src/pages/[...slug].tsx
```diff
@@ -458,10 +476,18 @@ export const getStaticProps: GetStaticProps = async ({ params }) => {
     const { slug } = params as IParams;
     const contents: ContentItem[] = await getAllContentItems();
 
-    //const sideNav = constructSideNav(primaryTag)
+    //TODO - not sure how to construct side nav for podcasts and videos since they don't have primary tags
```
Comment (mcolella14) [inline src/pages/[...slug].tsx:479]: Yeah I think we'd need some logic to parse the primary tag out of the tags prop and then slugify that. Or just try to pass the primary tag as a prop as well?

mcolella14 (approved):

PR #80, opened 2022-05-09
bug fix  - appending content type to existing other tags instead of creating new object.

harikakotipalli:


src/service/get-all-podcasts.ts
```diff
@@ -5,12 +5,20 @@ import getAllPodcastsFromAPI from '../api-requests/get-all-podcasts';
 export const getAllPodcasts = async (): Promise<Podcast[]> => {
     const podcasts = await getAllPodcastsFromAPI(STRAPI_CLIENT);
     podcasts.forEach(p => {
-        p.otherTags = {
-            contentType: {
+        if (p.otherTags) {
```
harikakotipalli, line 8: add content type to existing other tags for podcasts

src/service/get-all-videos.ts
```diff
@@ -5,12 +5,19 @@ import getAllVideosFromAPI from '../api-requests/get-all-videos';
 export const getAllVideos = async (): Promise<Video[]> => {
     const videos = await getAllVideosFromAPI(STRAPI_CLIENT);
     videos.forEach(v => {
-        v.otherTags = {
-            contentType: {
+        if (v.otherTags) {
```
harikakotipalli, line 8: add content type to existing other tags for videos instead of creating fresh object

mcolella14 (approved):

PR #82, opened 2022-05-09
[DEVHUB-1076] Add pipeline for content refresh

ajhenri:
Updates for DEVHUB-1076. This will be used for refresh published content, while also maintaining the latest revision that is currently live (either by promotion or rollback).

- Consolidate commits from older branch.
- Move out staging deploy and production deploy to separate manual steps that use drone CLI
- Publish and deploy based on commit hash that was last promoted

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -10,58 +10,97 @@ resources:
     memory: 4GiB
 
 steps:
-- name: publish-staging
-  image: plugins/kaniko-ecr
-  environment:
-    NPM_AUTH:
-      from_secret: NPM_AUTH
-    NPM_EMAIL:
-      from_secret: NPM_EMAIL
-    STRAPI_URL:
-      from_secret: STRAPI_URL_STAGING
-  settings:
-    create_repository: true
-    enable_cache: true
-    registry: 795250896452.dkr.ecr.us-east-1.amazonaws.com
-    repo: devrel/${DRONE_REPO_NAME}
-    build_args:
-      - NPM_AUTH
-      - NPM_EMAIL
-      - STRAPI_URL
-    tags:
-    - git-${DRONE_COMMIT_SHA:0:7}
-    - latest
-    access_key:
-      from_secret: ecr_access_key
-    secret_key:
-      from_secret: ecr_secret_key
-    values_files: [ "environments/staging.yaml" ]
-  when:
-    event:
-    - push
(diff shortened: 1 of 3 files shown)


.drone.yml
```diff
@@ -118,23 +158,153 @@ type: kubernetes
...
+    environment:
+      DRONE_SERVER: ${DRONE_SYSTEM_PROTO}://${DRONE_SYSTEM_HOST}
+      DRONE_TOKEN:
+        from_secret: drone_token
+
+trigger:
+  branch:
+    - main
+  event:
+    - custom
```
ajhenri, line 250: This will only run on `custom` event type, which is what the API will trigger.

.drone.yml
```diff
@@ -118,23 +158,153 @@ type: kubernetes
...
+  - name: checkout-to-current-release
+    image: alpine/git
+    environment:
+      APP_RELEASE:
+        from_secret: app_release_staging
+    commands:
+    # "$${APP_RELEASE}" is the latest staging revision that was promoted successfully.
+    - git checkout "$${APP_RELEASE}"
+
+  - name: publish-staging-from-trigger
```
ajhenri, line 208: This block shows that 1) `app_release_staging` is the SHA used for the following `publish-staging-from-trigger` build. The following `deploy-staging-from-trigger` step will use the build number produced from the prior step.

.drone.yml
ajhenri: I'm assuming there will be only one webhook, but I can add additional configuration. Currently, it's only the environment, but I think it makes sense to include another parameter to check.

.drone.yml
```diff
@@ -10,59 +10,98 @@ resources:
...
+      NPM_AUTH:
+        from_secret: NPM_AUTH
+      NPM_EMAIL:
+        from_secret: NPM_EMAIL
+      STRAPI_URL:
+        from_secret: STRAPI_URL_STAGING
+    settings:
+      create_repository: true
+      enable_cache: true
+      registry: 795250896452.dkr.ecr.us-east-1.amazonaws.com
```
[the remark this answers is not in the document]
ajhenri: There isn't a straightforward way to put this into an environment variable. I will revisit this in another PR, as this is already there in main.

ajhenri:
Thanks!

PR #84, opened 2022-05-10
L0 Pages

mcolella14:
[Jira Ticket ](https://jira.mongodb.org/browse/DEVHUB-1083)

The routes will be `/developer/languages`, `/developer/products`, and `/developer/technologies`.

These are all very similar pages, but making a template seemed like more work than it was worth at this point.

I did refactor a service that was returning just unique slugs of all tags. It now returns all unique tag objects, from which slugs (and other attributes) can be extracted. I needed this functionality for getting all unique products/technologies/languages.

The `logoPaths.ts` file is copied directly from Flora. I have to update and expose this in a Flora contribution for us to use it, at which point we can get rid of this file.


ajhenri:
Following up on this, realized the issue had to do with what's in the `main` branch and not with your PR. Also, as discussed yesterday, this is only affecting local development (`yarn dev`).

src/pages/languages.tsx
```diff
@@ -0,0 +1,235 @@
...
+const cSharpFeaturedConfig = {
+    imageConfig: {
+        src: 'https://webimages.mongodb.com/_com_assets/cms/ks67l72x98g3whceo-lang_c_sharp.svg?auto=format%252Ccompress',
+        variant: ESingleImageVariant.NO_RATIO,
+    },
+    cta: {
+        type: 'link-arrow' as CTAType,
+        text: 'Explore Content',
+        config: {
+            href: '/developer/languages/c-sharp',
```
ajhenri, line 68: This should be `/developer/languages/csharp`, as currently it goes to a 404

mcolella14:
Fixed some edge cases (including the one mentioned by @ajhenri ). I also noticed that we were still using the capitalized slug as the title on the L1 pages, so I fixed that.

mcolella14:
> LGTM. I noticed that "AWS" is shown as "Aws" and "GCP" is "Gcp" on the topic page. Not sure if that has to do with this PR or not, but something that I saw when testing.

Ah yeah so I noticed that too, and resolved it in my last commit.

PR #86, opened 2022-05-10
[bug fix] Tutorials and code examples pages not building w/ prod data

mcolella14:
Code Examples was failing because it had no content for the featured section. The featured section shouldn't render if it doesn't have enough content, so I enforced that.

Tutorials weren't rendering because, when we fail to map the tag to its corresponding flora icon, it yields undefined. Normally that's fine, but that step happens in `getStaticProps`, and undefined can't be serialized to json and passed to the pages. To solve, I just made the value `null` if its undefined, and handled the null case downstream in the template.


src/pages/code-examples.tsx
```diff
@@ -112,7 +112,7 @@ export const getStaticProps: GetStaticProps = async () => {
                     text: tag.name,
                     url: tag.slug + '/code-examples',
                 },
-                imageString: languageToLogo[tag.name],
+                imageString: languageToLogo[tag.name] || null,
```
harikakotipalli, line 115: why is the logic different for code-examples and tutorials alone?

harikakotipalli: also lot of code looks similar between code-examples and tutorials, can we please move that into common functionality ?

harikakotipalli: Also not in this PR , but why are we using mock data for videos?

mcolella14: That's a good point, it should be separated. I will make that change. And for the videos, i think that was just a mistake on my end, forgot to add the logic for getting the first 3 content pieces.

src/page-templates/content-type/utils.ts
```diff
@@ -2,7 +2,120 @@ import { PillCategory } from '../../types/pill-category';
...
+        languageTags.forEach(langTag => {
+            const existingTag = languages.find(
+                tag => tag.name === langTag.name
+            );
+            if (existingTag) {
+                existingTag.count += 1;
+            } else {
+                languages.push({ ...langTag, count: 1 });
+            }
+        });
```
ajhenri, line 42: This code is repeated for `techTags` and `productTags` in addition to this. It can be moved out into a utility method (including the sorting part below) for ease of use and to reduce the repetition here.

ajhenri (commented):
My comment is not blocking (can be done later if needed) and is similar to what Harika mentioned about code repetition. Aside from Harika's comments, this looks good to me.

ajhenri (approved):

harikakotipalli (approved):

PR #91, opened 2022-05-11
Devhub 1044 - Author Pages

harikakotipalli:
Author pages.

https://devcenter-ad43ag220-harikakotipalli.vercel.app/developer/author/ado-kukic


__tests__/requests/get-articles.test.ts
```diff
@@ -5,7 +5,7 @@ import { getArticles } from '../../src/api-requests/get-articles';
 
 const expectedQuery = gql`
     query Articles {
-        articles @rest(type: "Article", path: "/articles") {
+        articles @rest(type: "Article", path: "/articles?_limit=-1") {
```
ajhenri, line 8: Good catch on these!

src/pages/index.tsx
```diff
@@ -28,7 +31,8 @@ export default Home;
 
 export const getStaticProps: GetStaticProps = async ({}) => {
     const client = clientFactory('ApolloREST', process.env.STRAPI_URL);
-    const articles = await getArticles(client);
+    const articles = await getAllArticlesFromAPI(client);
+    console.log(articles.length);
```
ajhenri, line 35: console.log here should be removed

ajhenri (approved):
LGTM

__tests__/requests/get-articles.test.ts
ajhenri: Will we ever have a situation where we get too many articles for the site to handle? If this is the case, we should cap it anyway at say, 10k? Though this would probably be unlikely.

harikakotipalli: highly unlikely from strapi.

src/pages/author/[...slug].tsx
```diff
@@ -0,0 +1,311 @@
...
+                >
+                    <Eyebrow sx={eyebrowStyles}>Developer Topics</Eyebrow>
+                    <div sx={middleSectionStyles}>
+                        <SpeakerLockup
+                            sx={speakerStyles}
+                            profileImage={{
+                                src: image
+                                    ? image.url
+                                    : getPlaceHolderImage(undefined),
+                                alt: 'Random people avatar',
```
Comment (mcolella14) [inline src/pages/author/[...slug].tsx:123]: This should be `image.alt`, right?

mcolella14 (approved):
Looks great! Just the one nit about the `alt`.

PR #94, opened 2022-05-12
Gregc/all topics page

gfunk1230:
https://user-images.githubusercontent.com/6556242/168684499-25b9bd2f-ed28-4ef5-b3ed-82969be0a40c.mov


src/pages/topics.tsx
```diff
@@ -0,0 +1,226 @@
...
+            <GridLayout
+                sx={{
+                    paddingLeft: [
+                        'inc40',
+                        'inc40',
+                        'inc50',
+                        'inc70',
+                        'inc70',
+                        0,
+                    ],
```
mcolella14, line 99: Just a nit and doesn't really matter, but you can use `null` is these arrays to allow the value to carry over from the last one. For example `['inc40', 'inc40', 'inc50', 'inc70', 'inc70', 0]` could be `['inc40', null, 'inc50', 'inc70', null, 0]`. It's not too helpful in this use case, but for longer expressions it's kinda nice.

src/pages/topics.tsx
```diff
@@ -0,0 +1,226 @@
...
+                </div>
+            </GridLayout>
+        </>
+    );
+};
+
+export default Topic;
+
+export const getStaticProps: GetStaticProps = async ({}) => {
+    const client = clientFactory('ApolloREST', process.env.STRAPI_URL);
```
mcolella14, line 221: Don't think we need to bother instantiating this here.

mcolella14 (approved):
Looks good, only thing is that it looks like there is a bigger bottom margin on the designs.
Screen Shot 2022-05-17 at 9 39 14 AM
Screen Shot 2022-05-17 at 9 39 22 AM

I think you can use the `pageWrapper` style object to wrap the main content section of this page (https://github.com/mongodb/devcenter/blob/main/src/styled/styles.ts)

ajhenri (approved):

src/pages/topics.tsx
gfunk1230: Will remove in separate PR

PR #95, opened 2022-05-12
Search Page

mcolella14:
This search component shares a bit of logic with the content type aggregate pages, but the behavior of adjusting the URL based on the filters and search string make it difficult to break it into reusable logic. Given the time constraints I think we can optimize later.

This page renders a little slowly, and I checked and confirmed that this is an issue with flora. All of the flora components are wrapped in a `ThemeProvider`, which will render regardless of whether we memoize the component. I'm talking with Angelo about how to resolve this. To compensate, I added debouncing to the search box typing, which lightens the load on the browser to render everything.

I also decided to populate our search filters from the search collection itself via the realm function we have. Once we have separate staging and prod search collections, this realm function URL should be kept as an environment variable.

The page is `/developer/search`.


src/api-requests/get-all-search-content.ts
```diff
@@ -0,0 +1,13 @@
+import { SearchItem } from '../components/search/types';
+const searchEndpoint =
+    'https://data.mongodb-api.com/app/devhub-search-service-fldiy/endpoint/search_devcenter';
+
+// Should find a way to cache this response. Only use it in a few places for search filters but still.
```
ajhenri, line 5: I think it might be worth capturing this in a Jira ticket to be worked on post-launch.

src/components/search-filters/filter-group.tsx
```diff
@@ -11,168 +11,184 @@ import theme from '@mdb/flora/theme';
...
-                {filterString}
-            </TypographyScale>
-        );
-    })();
+            const filterString =
+                filterNames.length < 5
+                    ? filterNames.join(', ')
+                    : `${filterNames.slice(0, 5).join(', ')} +${
+                          filterNames.length - 5
+                      }`;
```
ajhenri, line 75: What do you think about moving 5 into a constant, something like `FILTER_STEP`? Then you can use it here and in the next few places (lines 106 - 186).

src/components/search/search.tsx
```diff
@@ -48,6 +49,18 @@ const Search: React.FunctionComponent<SearchProps> = ({
         setSearchString(event.target.value);
     };
 
+    const debouncedOnSearch = useMemo(
+        () => debounce(onSearch, 400), // Not sure what this value should be.
```
ajhenri, line 53: I think 400 is a good starting point.

src/page-templates/content-type/utils.ts
```diff
@@ -156,10 +158,49 @@ export const getFilters = async (contentType: PillCategory) => {
                         type: l1.type,
                         name: l1.name,
                         subItems: [l2FilterItem],
-                        count: category === contentType ? 1 : 0,
+                        count: allFilters || type === contentType ? 1 : 0,
                     };
                     return filterItems.push(l1FilterItem);
                 }
+            } else if (tag.type === 'CodeLevel') {
+                // Basically repeating the logic for L2s, can probably be separated and reused.
```
ajhenri, line 166: Similar to the comment above regarding optimization, I would capture this work in a ticket to revisit.

ajhenri (commented):
Thanks for working on this. Looks tough to do in this amount of time.

I noticed something on the Vercel deploy: some tags are showing up with an empty block.
Screen Shot 2022-05-12 at 6 34 20 PM

Otherwise, I have not noticed any other issues besides what you have already mentioned regarding optimization and code repetition. I will review this again soon but wanted to share initial feedback.

mcolella14:
> Thanks for working on this. Looks tough to do in this amount of time.
> 
> I noticed something on the Vercel deploy: some tags are showing up with an empty block. Screen Shot 2022-05-12 at 6 34 20 PM
> 
> Otherwise, I have not noticed any other issues besides what you have already mentioned regarding optimization and code repetition. I will review this again soon but wanted to share initial feedback.

Ah yeah I'm actually tackling that in a separate PR right now

src/components/search-filters/filter-group.tsx
```diff
@@ -11,168 +11,184 @@ import theme from '@mdb/flora/theme';
...
-            {expanded && (
-                <div>
-                    <div sx={itemsStyles(title)}>
-                        {items
-                            .slice(0, showAll ? undefined : 5) // Show 5 to start, then all if they click "Show more"
-                            .map(filter => {
-                                if (filter.subItems && filter.subItems.length) {
+                )}
+                {expanded && (
+                    <div>
```
gfunk1230, line 108: Do we need this `div` here or can we swap it out for a fragment `<>`?

src/components/search-filters/filter-group.tsx
```diff
@@ -11,168 +11,184 @@ import theme from '@mdb/flora/theme';
...
+                        <div sx={itemsStyles(title)}>
+                            {items
+                                .slice(0, showAll ? undefined : 5) // Show 5 to start, then all if they click "Show more"
+                                .map(filter => {
```
gfunk1230, line 112: Might be worth destructuring `filter` so that you don't need to reference it in multiple places.

src/components/search/search.tsx
gfunk1230: I normally do 800-1000 just bc 400 is too subtle but if you want to keep 400 it's fine.

gfunk1230 (approved):
minor comments but LGTM!

src/components/search-filters/filter-group.tsx
mcolella14: So my intention was to have the filters and the "Show More" link to be enclosed in the same element in the DOM, but thinking about it now, it might make more sense to have them as siblings, so I'll put them in a fragment.

ajhenri (approved):

PR #97, opened 2022-05-13
Mc/misc

mcolella14:
This covers various bug fixes and missing pieces:

- Added the Code Level (Full Application/Snippet) to cards and as a filter on Topic->Code Examples page.
- Fix bug where a blank tag would show, and be explicit about what tags we want to show in the tag section component.
- Encoded the query params when hitting our realm function.
- On Topic Content Type and Content pages, there is supposed to be a top item on the side nav representing the primary tag, and that's supposed to be a clickable link. That has now been implemented.
- Implemented breadcrumbs in every place they show up in the designs.
- On the Topic Content Page for L1s (e.g. `/products/atlas/articles`), the subitems were not being filtered by which ones have the correct content type. For example, we don't want to sho Full Text Search on that page if there are no articles about Full Text Search (clicking on that link would result in a 404 since no page is built for a topic without content). I filtered these and added the correct links to the subitems.

### DIFF
--- a/src/components/breadcrumbs/utils.ts
+++ b/src/components/breadcrumbs/utils.ts
@@ -1,0 +1,34 @@
+import { Crumb } from './types';
+import { getMetaInfoForTopic } from '../../service/get-meta-info-for-topic';
+
+const categoryToCrumb: { [key: string]: Crumb } = {
+    products: { text: 'Products', url: '/developer/products' },
+    languages: { text: 'Languages', url: '/developer/languages' },
+    technologies: { text: 'Technologies', url: '/developer/technologies' },
+};
+
+export const getBreadcrumbsFromSlug = async (
+    slug: string
+): Promise<Crumb[]> => {
+    const crumbs: Crumb[] = [
+        { text: 'MongoDB Developer Center', url: '/' },
+        { text: 'Developer Topics', url: '/developer/topics' },
+    ];
+    const slugList = slug.startsWith('/')
+        ? slug.split('/').slice(1)
+        : slug.split('/');
+
+    const categoryCrumb = categoryToCrumb[slugList[0]];
+    if (categoryCrumb) {
+        crumbs.push(categoryCrumb);
+    }
+
+    for (let i = 2; i < slugList.length; i++) {
+        const crumbSlug = '/' + slugList.slice(0, i).join('/');
+        const metaInfoForTopic = await getMetaInfoForTopic(crumbSlug);
+        if (metaInfoForTopic?.tagName) {
+            crumbs.push({ text: metaInfoForTopic.tagName, url: crumbSlug });
+        }
+    }
+    return crumbs;
+};
--- a/src/components/breadcrumbs/utils.ts
+++ b/src/components/breadcrumbs/utils.ts
@@ -2,9 +2,9 @@ import { Crumb } from './types';
 import { getMetaInfoForTopic } from '../../service/get-meta-info-for-topic';
 
 const categoryToCrumb: { [key: string]: Crumb } = {
-    products: { text: 'Products', url: '/developer/products' },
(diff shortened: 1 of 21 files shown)


src/components/breadcrumbs/utils.ts
```diff
@@ -0,0 +1,34 @@
+import { Crumb } from './types';
+import { getMetaInfoForTopic } from '../../service/get-meta-info-for-topic';
+
+const categoryToCrumb: { [key: string]: Crumb } = {
+    products: { text: 'Products', url: '/developer/products' },
```
gfunk1230, line 5: Not 100 percent sure about this one but I thought we can use Amanda's utility function `getURLPath` and not reference developer in the URLs?

src/pages/[...slug].tsx
```diff
@@ -435,22 +461,22 @@ interface IParams extends ParsedUrlQuery {
...
 } // Need this to avoid TS errors.
 
-const removesErroringArticles = (contents: ContentItem[]) => {
-    const removeItems = [
-        'PyMongoArrow: Bridging the Gap Between MongoDB and Your Data Analysis App',
-        'new-time-series-collections',
-        'mongodb-charts-embedding-sdk-react/',
-        'build-movie-search-application',
-    ];
+// const removesErroringArticles = (contents: ContentItem[]) => {
```
Comment (gfunk1230) [inline src/pages/[...slug].tsx:464]: Did we want to keep this in here?

gfunk1230 (approved):
Minor comments but LGTM!

src/components/breadcrumbs/utils.ts
ajhenri: Yeah, I think for now that can be used here to avoid referencing /developer.

src/components/icons/BaseIcon.tsx
```diff
@@ -1,15 +1,17 @@
...
 const BaseIcon: React.FunctionComponent<BaseIconProps> = ({
     className,
     children,
+    isFilled = false,
 }) => {
     return (
         <svg
             viewBox="0 0 24 24"
-            fill="none"
+            {...(isFilled ? {} : { fill: 'none' })}
```
ajhenri, line 14: Nit: I don't think the empty object is needed. I _think_ you can instead do `...( isFilled && { fill: 'none' } )` and it would work the same?

src/pages/[l1_l2]/[topic]/[...slug].tsx
```diff
@@ -238,6 +246,8 @@ export const getStaticProps: GetStaticProps = async ({ params }) => {
     */
 
     const pathComponents = [l1_l2, topic].concat(slug);
+
+    // const crumbs = await getBreadcrumbsFromSlug(pathComponents.join('/'));
```
Comment (ajhenri) [inline src/pages/[l1_l2]/[topic]/[...slug].tsx:250]: Just noting that this line is commented out. Do we not need it?

ajhenri (approved):
LGTM, aside from Greg's comments. I had a couple of comments but none of them are blocking this PR.

src/pages/[...slug].tsx
Comment (mcolella14) [reply src/pages/[...slug].tsx:464]: Yeah it may be good to have it in case we have any other content items that give us trouble and we need to quickly disable them.

src/components/icons/BaseIcon.tsx
mcolella14: Ah yeah looks like that works, didn't know that, thank you!

src/pages/[l1_l2]/[topic]/[...slug].tsx
Comment (mcolella14) [reply src/pages/[l1_l2]/[topic]/[...slug].tsx:250]: I think I left it in there in case we need to put breadcrumbs on this page (they're not in the designs now), but if that ever happens it won't be hard to implement, so I'll take this out.

PR #105, opened 2022-05-17
[DEVHUB-1111] Added GTM to NextJS

gfunk1230:
Added GTM to NextJS App.

### DIFF
--- a/src/pages/_app.tsx
+++ b/src/pages/_app.tsx
@@ -1,11 +1,25 @@
+import { useEffect } from 'react';
+import Script from 'next/script';
+import { useRouter } from 'next/router';
 import type { AppProps } from 'next/app';
 import Head from 'next/head';
 import theme from '@mdb/flora/theme';
 import { ThemeProvider } from '@theme-ui/core';
-
+import { GTM_ID, pageView } from '../lib/gtm';
 import Layout from '../components/layout';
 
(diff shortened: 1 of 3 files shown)


src/pages/_document.tsx
```diff
@@ -0,0 +1,24 @@
+import Document, { Html, Head, Main, NextScript } from 'next/document';
```
ajhenri, line 1: I did not know about next/document. This can be used to alter what's in the body tag. Neat!

ajhenri (approved):
LGTM

src/pages/_app.tsx
```diff
@@ -1,18 +1,46 @@
...
 import theme from '@mdb/flora/theme';
 import { ThemeProvider } from '@theme-ui/core';
-
+import { GTM_ID, pageView } from '../lib/gtm';
 import Layout from '../components/layout';
 
 function MyApp({ Component, pageProps }: AppProps) {
+    {
+        /* Is used to track route changes and send page views to Google Tag Manager */
+    }
```
ajhenri, line 14: @gfunk1230 I approved but wanted to point out that the brackets aren't needed here.

gfunk1230: Good catch! will remove!

mcolella14 (approved):

PR #106, opened 2022-05-17
New product logos, various bug fixes.

mcolella14:
The main change was implementing the new Flora logos for technologies/languages.

We were ending up with a lot of 404s because all the `metaInfo` stuff was not filtering out tags that didn't have content associated with it, which is of no value to us. I updated `getAllMetaInfo` to filter out these orphaned tags.

There was also a bug with the search checkboxes, and some weird behavior with the images/icons when we can't find the right one.


harikakotipalli (approved):

harikakotipalli:
LGTM! My idea was to have meta info have everything and then we use content data on our pages to map to this info, while constructing. either ways will work for me.

PR #107, opened 2022-05-18
Hk/misc

harikakotipalli:
Adding retry link for network errors. I tested 10 times and it worked fine. If it fails further we have to change configs and try again.

https://www.apollographql.com/docs/react/api/link/apollo-link-retry/

Fixed issue with getPodcasts and getVideos which is preventing yarn dev to run.


mcolella14 (approved):

ajhenri (approved):
LGTM

PR #115, opened 2022-05-20
Fix regexes for old devhub js/img/json file rewrites

ajhenri:
Noticed an issue with the regexes used to pull the js/img/json files from the old site and it is causing 404s on pages for nextjs, or anything ending in js. 

Once the production bucket is refreshed with the latest files, I'll make these more specific (e.g. per actual file).


mcolella14 (approved):

ajhenri:
Thanks!

PR #122, opened 2022-05-20
Feature/ssr

harikakotipalli:
this potentially reduces build time , since authors pages will be serverside rendered. I'm not sure we go this route or not, so I just commented the getStaticPaths stuff in case we want to go back to static site rendering on these pages.


mcolella14 (commented):
Won't these pages take a super long time to load doing SSR since we have to wait for a response from both the authors and articles endpoints?

harikakotipalli:
> 

only first the first author. All the subsequent author pages are using the already called GetAllContent and GetAllAuthors.
You can test it.

harikakotipalli:
> @mcolella14  only first the first author. All the subsequent author pages are using the already called GetAllContent and GetAllAuthors.
> You can test it.

ajhenri:
This looks awesome!
We might want to take advantage of ssr-caching https://nextjs.org/docs/going-to-production#caching and play around with the cache-control headers (later on). 

> Won't these pages take a super long time to load doing SSR since we have to wait for a response from both the authors and articles endpoints?
Is it possible to not load all articles on authors page? I recall you both talking about some limitation regarding this.

ajhenri (commented):
This looks awesome! Thanks for working on this.
We might want to take advantage of ssr-caching https://nextjs.org/docs/going-to-production#caching and play around with the cache-control headers (later on).

> Won't these pages take a super long time to load doing SSR since we have to wait for a response from both the authors and articles endpoints?

Is it possible to not load all articles on authors page? I recall you both talking about some limitation regarding this.

mcolella14:
> > @mcolella14  only first the first author. All the subsequent author pages are using the already called GetAllContent and GetAllAuthors.
> 
> > You can test it - go to an article - click on author - this first author might take time. Now go to another article click on author - this should be fast.

Ah that's awesome! I assumed next wouldn't maintain that Apollo cache across requests, very cool that it does.

PR #128, opened 2022-05-23
SEO Fixes

ajhenri:
- Fix issue with canonical on [...slug] route.
- Use trailing slashes, similar to former Developer Hub. This should not affect anything, as this behavior was previously by default redirecting trailing slash paths to be without the trailing slash. Now the behavior will be the opposite.

Example (External Canonical)
https://devcenter-4szrhlbpi-harikakotipalli.vercel.app/developer/products/realm/realm-kotlin-0-6-0/

### DIFF
--- a/src/pages/[...slug].tsx
+++ b/src/pages/[...slug].tsx
@@ -1,5 +1,7 @@
 import type { GetStaticProps, NextPage } from 'next';
 import { NextSeo } from 'next-seo';
+import { useRouter } from 'next/router';
+import getConfig from 'next/config';
 import { ParsedUrlQuery } from 'querystring';
 import Image from 'next/image';
 import React, { useState } from 'react';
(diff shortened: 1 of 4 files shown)


src/pages/[...slug].tsx
```diff
@@ -433,6 +437,9 @@ const ContentPage: NextPage<ContentPageProps> = ({
         handle: seo?.twitter_creator,
         site: seo?.twitter_site,
     };
+    let canonicalUrl = seo?.canonical_url
+        ? seo?.canonical_url
+        : publicRuntimeConfig.absoluteBasePath + router.asPath;
```
Comment (harikakotipalli) [inline src/pages/[...slug].tsx:442]: this is believe is self referencing if they don't explicitly add any SEO in strapi.

Comment (ajhenri) [reply src/pages/[...slug].tsx:442]: Correct, `publicRuntimeConfig.absoluteBasePath + router.asPath` will be the existing page (self referencing value).

harikakotipalli (approved):

ajhenri:
thanks @harikakotipalli !

PR #135, opened 2022-05-24
Last minute SEO requests batch

ajhenri:
Tickets: 
- [DEVHUB-1224](https://jira.mongodb.org/browse/DEVHUB-1224)
- [DEVHUB-1244](https://jira.mongodb.org/browse/DEVHUB-1244)
- [DEVHUB-1246](https://jira.mongodb.org/browse/DEVHUB-1246)

Regarding DEVHUB-1244, Next.js doesn't give us the ability to use a trailing slash for a _specific_ page. It has to be either for all pages or for no pages. Elizabeth said for SEO we need to have a trailing slash specifically for the homepage, but the only way I can see to do that is this.


src/pages/search.tsx
```diff
@@ -450,7 +450,10 @@ const Search: NextPage<SearchProps> = ({
 
     return (
         <>
-            <NextSeo title={'Search | MongoDB'} />
+            <NextSeo
+                title={'Search | MongoDB'}
+                noindex={router.asPath === '/search/' ? false : true}
```
mcolella14, line 455: Will this prevent it from indexing the search page itself?

mcolella14 (approved):

src/pages/search.tsx
ajhenri: @mcolella14  No actually, `noindex` is only true if the search page has parameters. Otherwise, if the path value is `/search/` (landing page) then noindex would be false and page is indexed.

PR #145, opened 2022-05-25
removing load more functionality from author pages to make content discoverable

harikakotipalli:
in reference to https://mongodb.slack.com/archives/C02NFHPSZ6Z/p1653491029496949

### DIFF
--- a/src/pages/author/[...slug].tsx
+++ b/src/pages/author/[...slug].tsx
@@ -82,23 +82,23 @@ const AuthorPage: NextPage<AuthorPageProps> = ({
     const titleAndLocation = getTitleAndLocation(title, location);
     const hasSocial = facebook || twitter || linkedin || youtube;
 
-    const LENGTH = articles.length;
-    const DATA = articles;
-    const LIMIT = 8;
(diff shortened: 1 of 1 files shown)


src/pages/author/[...slug].tsx
```diff
@@ -82,23 +82,23 @@ const AuthorPage: NextPage<AuthorPageProps> = ({
...
-
-    const loadMore = () => {
-        const newIndex = index + LIMIT;
-        const newShowMore = newIndex <= LENGTH - 1;
-        const newList = list.concat(DATA.slice(index, newIndex));
-        setIndex(newIndex);
-        setList(newList);
-        setShowMore(newShowMore);
-    };
+    // const LENGTH = articles.length;
```
Comment (gfunk1230) [inline src/pages/author/[...slug].tsx:85]: I assume this is all temporary so that's why we are just commenting it out for now?

gfunk1230 (approved):

src/pages/author/[...slug].tsx
Comment (harikakotipalli) [reply src/pages/author/[...slug].tsx:85]: yes it is. they want to avoid load more for seo purposes.

PR #148, opened 2022-05-25
RSS Feed, author lockup image and card background.

mcolella14:
Jira Ticket:

[DEVHUB-1250](https://jira.mongodb.org/browse/DEVHUB-1250)

Main thing is the RSS Feed. I originally had it running on `getStaticProps` of the homepage, but according to Next: 

> Only assets that are in the public directory at [build time](https://nextjs.org/docs/api-reference/cli#build) will be served by Next.js. Files added at runtime won't be available. We recommend using a third party service like [AWS S3](https://aws.amazon.com/s3/) for persistent file storage.

The issue with doing it outside of the context of the app is that I couldn't it to play well with using modules in the app (like `getAllContentItems`). So I decided to just pull from the search Realm function since it should be the same and only requires a single API request. I put it in the webpack config of the Next config so that it runs before build and can add `rss.xml` to the `public` directory and have that be present when the app is built. (see https://github.com/vercel/next.js/discussions/17479#discussioncomment-434551).

The feed will be available at the same url as the existing DevHub (`/developer/rss.xml`).

Also made some minor design fixes that I noticed.

### DIFF
--- a/src/scripts/build-rss-feed.js
+++ b/src/scripts/build-rss-feed.js
@@ -1,0 +1,41 @@
+const Feed = require('feed').Feed;
+const fs = require('fs');
+const axios = require('axios').default;
+
+async function buildRssFeed(baseUrl) {
+    const date = new Date();
+
+    const feed = new Feed({
+        title: 'MongoDB Developer Center',
+        description: 'MongoDB Developer Center RSS Feed',
+        id: baseUrl,
+        link: baseUrl,
+        language: 'en',
+        updated: date,
+        generator: 'Next.js using Feed for Node.js',
+        feedLinks: {
+            rss2: `${baseUrl}/rss.xml`,
+        },
+    });
+
+    const searchEndpoint =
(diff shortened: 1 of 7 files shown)


src/scripts/build-rss-feed.js
```diff
@@ -0,0 +1,41 @@
...
+        updated: date,
+        generator: 'Next.js using Feed for Node.js',
+        feedLinks: {
+            rss2: `${baseUrl}/rss.xml`,
+        },
+    });
+
+    const searchEndpoint =
+        'https://data.mongodb-api.com/app/devhub-search-service-fldiy/endpoint/search_devcenter?s=';
+    const posts = await axios.get(searchEndpoint);
```
harikakotipalli, line 23: Good call @mcolella14

harikakotipalli (approved):

harikakotipalli:
Thank you so much for taking this one @mcolella14

ajhenri (approved):

PR #151, opened 2022-05-26
Add thumbnail loader to all images that need it.

mcolella14:
The rating-star svg seems fine, think it might not affect SVGs.


ajhenri (approved):
LGTM, but check on Vercel

harikakotipalli (approved):

PR #161, opened 2022-05-26
Remove /learn redirect causing issues

ajhenri:
Remove /learn redirect which was causing issues testing CDN with Angelo.


mcolella14 (approved):

ajhenri:
Going to force this in due to the fact that this can't be tested anywhere other than the CDN.

PR #174, opened 2022-06-01
Create feedback on rate, and update on subsequent feedback.

mcolella14:
Jira Ticket: [DEVHUB-1267](https://jira.mongodb.org/browse/DEVHUB-1267)

I made 2 new realm functions shown in [this PR on the api service](https://github.com/10gen/devhub-api-service/pull/24). This will hit the `create_feedback` function when the user rates an article, receives the _id of that newly created document, and then the `update_feedback` function when they submit the checkbox or text feedback.


harikakotipalli (approved):

harikakotipalli:
I believe they added new fields, hence the changes.

mcolella14:
With the new rate limiting, I really want to test this in a staging environment because when I test locally I only have my IP, and can't really simulate how this would work when not locally run. Given that, I'm converting this to a draft so it is not included in the pre-World release.

.github/workflows/vercel-deploy.yml
```diff
@@ -21,5 +21,5 @@ jobs:
                   vercel-token: ${{secrets.VERCEL_TOKEN}}
                   vercel-org-id: ${{secrets.VERCEL_ORG_ID}}
                   vercel-project-id: ${{secrets.VERCEL_PROJECT_ID}}
-                  vercel-args: '--prod --build-env NPM_AUTH=${{ secrets.ARTIFACTORY_NPM_AUTH }} --build-env NPM_EMAIL=${{ secrets.ARTIFACTORY_NPM_EMAIL }} --build-env STRAPI_URL=${{ secrets.STRAPI_DEV_URL }}'
+                  vercel-args: '--prod --build-env NPM_AUTH=${{ secrets.ARTIFACTORY_NPM_AUTH }} --build-env NPM_EMAIL=${{ secrets.ARTIFACTORY_NPM_EMAIL }} --build-env STRAPI_URL=${{ secrets.STRAPI_DEV_URL }} --build-env REALM_API_URL=${{ secrets.REALM_API_URL }} --build-env REALM_SEARCH_URL=${{ secrets.REALM_SEARCH_URL }}'
```
ajhenri, line 24: If you want this to work in Vercel, you might be able to do `--build-env HOST_URL=${{ secrets.VERCEL_URL }}`, to set the HOST_URL. Mentioning since I saw that the preview environment had a 500.

src/pages/_middleware.ts
```diff
@@ -1,9 +1,65 @@
...
+        pathname.startsWith('/api/') && ['POST', 'PUT'].includes(req.method);
+
+    if (checkRequest) {
+        let headers: { [key: string]: string } = {
+            'Content-Type': 'application/json',
+        };
+
+        // Rate limit
+        try {
+            await limiter.check(headers, MAX_FEEDBACK_PER_PERIOD, req.ip || '');
```
ajhenri, line 28: Nice!

src/pages/_middleware.ts
```diff
@@ -1,9 +1,65 @@
...
+                    status: 500,
+                    headers,
+                }
+            );
+        }
+
+        // Minimal bot detection by checking the user agent for real browser info.
+        if (
+            !req.ua?.browser?.name ||
+            req.nextUrl.origin !== process.env.HOST_URL
```
ajhenri, line 44: req.nextUrl.origin would have `http`/`https` prepended so the comparison would fail. The `HOST_URL` doesn't contain the protocol.

ajhenri (commented):
Sorry for the delay on this due to being at World. This is great work! Thanks a lot for making these changes. I don't want to ask you to do too much more than this, since we plan to rebuild things down the line.

Really like the rate limiting added here. We will definitely be using this in the future. I added a few early comments on some things I noticed.

.github/workflows/vercel-deploy.yml
mcolella14: ahhh ok that makes sense, I was just testing on Vercel and was confused about this.

ajhenri: Ah that's unfortunate. Kind of ugly, but you can do something like https://github.com/mongodb/devcenter/blob/main/next.config.js#L5.

mcolella14: Ahh I see, I didn't realize `VERCEL_URL` was already populated by Vercel's deployment, but I think the logic we do in the config should work!

mcolella14:
Preview environment still has a 500, gonna wait until we can check the logs to continue.

mcolella14:
Closing to open another and hopefully refresh the environment variables passed to Vercel.

PR #180, opened 2022-06-06
Add deploy instructions

ajhenri:
Let me know if there is anything else I can add here, or if further explanation is needed.


mcolella14 (approved):
LGTM, thank you for writing this up!

PR #181, opened 2022-06-06
mc/Miscellaneous Pre-World things.

mcolella14:
Jira Tickets:

- [DEVHUB-981](https://jira.mongodb.org/browse/DEVHUB-981)
- [DEVHUB-1259](https://jira.mongodb.org/browse/DEVHUB-1259)
- [DEVHUB-1276](https://jira.mongodb.org/browse/DEVHUB-1276)

### DIFF
--- a/src/components/social-buttons/utils.ts
+++ b/src/components/social-buttons/utils.ts
@@ -1,0 +1,61 @@
+import { Author } from '../../interfaces/author';
+import { Tag } from '../../interfaces/tag';
+
+export const getTweetText = (
+    authors: Author[],
+    heading: string,
+    tags: Tag[]
+): string => {
+    const twitterTextElements: string[] = [heading, '#MongoDB'];
+
+    const languages = tags.filter(tag => tag.type === 'ProgrammingLanguage');
+    if (languages.length) {
+        languages.forEach(({ name }) => {
+            let languageTag: string | null = null;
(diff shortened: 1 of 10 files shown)


src/components/social-buttons/utils.ts
```diff
@@ -0,0 +1,61 @@
...
+            } else if (name === 'C') {
+                languageTag = 'clang';
+            } else if (name === 'Swift') {
+                languageTag = 'swiftlang';
+            } else if (name === 'Go') {
+                languageTag = 'golang';
+            } else if (name === 'Rust') {
+                languageTag = 'rustlang';
+            } else if (name === 'C#') {
+                languageTag = 'csharp';
```
ajhenri, line 26: Not blocking for this PR, but do these need to change whenever the tag structure changes?

ajhenri (approved):
LGTM! Thanks for the quick updates

src/components/social-buttons/utils.ts
mcolella14: Yeah, I think ideally we'd have a field for the correct hashtag to use, but for now we have to hard-code this mapping. Anything that doesn't have special characters in it should automatically get a hashtag though with this logic, so it would only be for when we create/change a language tag that has a special character (or has a generic name like Rust or Go and needs a different hashtag).

PR #182, opened 2022-06-06
Hk/instant preview

harikakotipalli:


src/pages/api/preview.js
```diff
@@ -0,0 +1,4 @@
+export default function handler(req, res) {
+    res.setPreviewData({});
+    res.redirect('/developer/preview' + req.query.redirect);
+}
```
mcolella14, line 4: Are we ok with having unpublished articles publicly accessible through this preview route?

mcolella14 (commented):
Just the thing about public accessibility, but other than that everything looks good!

src/pages/api/preview.js
harikakotipalli: I'm trying to add a secret key, which is needed to be passed for preview to work. That secret key must be sent from Strapi and it will be stored in our env / secret variables.

src/pages/preview/[...slug].tsx
```diff
@@ -0,0 +1,67 @@
...
+            contentItem={contentItem}
+            tertiaryNavItems={tertiaryNavItems}
+            relatedContent={relatedContent}
+        />
+    );
+};
+
+export default ContentPage;
+
+export const getServerSideProps = async (context: any) => {
```
Comment (ajhenri) [inline src/pages/preview/[...slug].tsx:45]: Awesome 🎉

src/service/get-preview-content.ts
```diff
@@ -0,0 +1,17 @@
+import { ContentItem } from '../interfaces/content-item';
+import { getAllDraftArticlesFromAPI } from '../api-requests/get-articles';
+import { STRAPI_CLIENT } from '../config/api-client';
+import { mapArticlesToContentItems } from './get-all-content';
+import { Article } from '../interfaces/article';
+
+export const getPreviewContent: (
+    calculatedSlug: string
+) => Promise<ContentItem[]> = async calculatedSlug => {
+    console.log(calculatedSlug);
```
ajhenri, line 10: nit: console.log-- is this needed here?

src/pages/api/preview.js
```diff
@@ -0,0 +1,4 @@
+export default function handler(req, res) {
+    res.setPreviewData({});
+    res.redirect('/developer/preview' + req.query.redirect);
```
ajhenri, line 3: How does this work from the author's point of view?

ajhenri (commented):
This looks great, Harika. Once the secret key is added, I'm inclined to approve.

src/pages/preview/[...slug].tsx
Comment (harikakotipalli) [reply src/pages/preview/[...slug].tsx:51]: ideally we should just repurpose  [...slug].tsx but next.js is not resolving  to catch all [...slug].tsx , so had to make a new path /preview , may be we can explore later.

src/api-requests/get-articles.ts
```diff
@@ -114,3 +116,21 @@ export const getAllArticlesFromAPI = async (
...
+    calculatedSlug = '"' + calculatedSlug + '"';
+    const query = gql`
+        query Articles {
+            articles(_publicationState : "preview", calculated_slug : ${calculatedSlug})
+            @rest(type: "Article", path: "/new-articles?{args}") {
+                ${fields}
+            }
+        }`;
+    const { data }: ApolloQueryResult<{ articles: Article[] }> =
+        await client.query({ query, fetchPolicy: 'no-cache' });
```
harikakotipalli, line 133: fetchPolicy: 'no-cache' - will always make network call for this query. it is needed since the users want to see the changes in live and there is no room for stale data.

src/pages/preview/[...slug].tsx
```diff
@@ -36,20 +41,14 @@ const ContentPage: NextPage<ContentPageProps> = ({
...
             tertiaryNavItems={tertiaryNavItems}
             relatedContent={relatedContent}
+            previewMode={previewMode}
         />
     );
 };
 
 export default ContentPage;
 
 export const getServerSideProps = async (context: any) => {
```
Comment (mcolella14) [inline src/pages/preview/[...slug].tsx:51]: Should we be checking here for `context.preview`? I'm following along here https://nextjs.org/docs/advanced-features/preview-mode#step-2-update-getstaticprops. I may be misunderstanding, but it seems like anyone could pass in the slug of an unpublished article and see it by going to `developer/preview/<some_draft_slug>`?

mcolella14 (approved):
Looks great! Just the one thing about the preview pages route.

src/pages/preview/[...slug].tsx
Comment (harikakotipalli) [reply src/pages/preview/[...slug].tsx:51]: @mcolella14 true.Context.preview is always coming as true whether it's coming from redirect or url is directly from preview url, same with Context.previewData. I'm exploring how to have a route protected in next.js, but don't have luck yet.

Comment (mcolella14) [reply src/pages/preview/[...slug].tsx:51]: Have you tried going directly to the preview route from an incognito window? I think the preview mode sets a cookie on your browser, so it would make sense for `context.preview` to be `true` if you went to the preview page route after hitting the preview api route a while ago. If it's still true with an incognito window, then I have no idea lol.

PR #185, opened 2022-06-08
[DEVHUB-1101] Refactor search

mcolella14:
Mainly moved some files around, and create a search hook that encompasses most of the search logic.


src/components/search/empty-state.tsx
```diff
@@ -0,0 +1,29 @@
...
+const EmptyState: React.FunctionComponent = () => (
+    <div
+        sx={{
+            display: 'flex',
+            flexDirection: 'column',
+            justifyContent: 'center',
+            alignItems: 'center',
+        }}
+    >
+        <div>
```
gfunk1230, line 16: Why do we need this div here?

src/components/search/search.tsx
```diff
@@ -155,23 +90,29 @@ const Search: React.FunctionComponent<SearchProps> = ({
                         name="Snippet"
                         label="Code Snippets"
                         onToggle={checked => onCheckToggle(checked, 'Snippet')}
-                        checked={codeLevelFilters.includes('Snippet')}
+                        checked={
+                            !!allFilters.find(filt => filt.name === 'Snippet')
```
gfunk1230, line 94: small nitpick but I would have created a `const` called `isSnippetChecked` and reference it here.

src/components/search/search.tsx
```diff
@@ -155,23 +90,29 @@ const Search: React.FunctionComponent<SearchProps> = ({
...
                     />
                     <Checkbox
                         name="Full Application"
                         label="Full Applications"
                         onToggle={checked =>
                             onCheckToggle(checked, 'Full Application')
                         }
-                        checked={codeLevelFilters.includes('Full Application')}
+                        checked={
+                            !!allFilters.find(
```
gfunk1230, line 104: small nitpick but I would have created a `const` called `isFullApplicationChecked` and referenced it here.

gfunk1230 (approved):
Made some recommendations but LGTM!

src/components/search/empty-state.tsx
mcolella14: I think I was having trouble getting the `Image` component to actually center without the `div`. I think there may be some styling on that component making that difficult.

PR #188, opened 2022-06-09
Create feedback on rate, and update on subsequent feedback.

mcolella14:
Jira Ticket: [DEVHUB-1267](https://jira.mongodb.org/browse/DEVHUB-1267)

I made 2 new realm functions shown in https://github.com/10gen/devhub-api-service/pull/24. This will hit the create_feedback function when the user rates an article, receives the _id of that newly created document, and then the update_feedback function when they submit the checkbox or text feedback.

This is the second PR I've opened for this, hoping to resolve some Vercel issues.


.drone.yml
```diff
@@ -17,6 +17,10 @@ steps:
         from_secret: NPM_AUTH
       NPM_EMAIL:
         from_secret: NPM_EMAIL
+      REALM_API_URL:
```
ajhenri, line 20: I know I brought this up in https://github.com/10gen/devhub-api-service/pull/24, but we never followed up on it. Any reason why you're not using a token / secret to secure this API?

.drone.yml
```diff
@@ -17,6 +17,10 @@ steps:
...
+        from_secret: REALM_API_URL
+      REALM_SEARCH_URL:
```
ajhenri, line 22: For search, I didn't think we needed to include here, but this is fine. I don't think you need a secret key for this one.

.drone.yml
mcolella14: So from talking with Grant and researching, we decided that rate limiting was probably the most important thing, and that using a token wasn't super useful given the public nature of this functionality. And I know you mentioned that the [educators page](https://devcenter-frontend.devrel.staging.corp.mongodb.com/developer/academia/educators/) is using tokens, but that form is using the Realm SDK rather than making a call to an API, so I think it's a bit different in how it would function on our site.

.drone.yml
mcolella14: Ahh ok, I just figured we could keep our realm URLs obscured across the board with this.

ajhenri (approved):
LGTM

PR #199, opened 2022-06-15
Include origin in put/post headers

mcolella14:
`req.nextUrl.origin` is not actually the request origin apparently, just the origin of the next server ([which is always gonna be `loclhost` I guess?](https://github.com/vercel/next.js/blob/canary/packages/next/server/web/next-url.ts#L258)). Now we grab the origin from the browser and pass it as a header.


ajhenri (approved):
LGTM. If this ends up not working, I am ok with dropping the origin check.

PR #201, opened 2022-06-15
Take out user agent checks and move rate limiting

mcolella14:
So the user agent checks weren't working, not sure why, but they're [changing around the behavior of that attribute soon anyway](https://github.com/vercel/next.js/pull/37512). The `ip` attribute on the `NextRequest` was also not working, so I put the IP checking and rate limiting in the api routes themselves. This is less Ideal, but it might work.

### DIFF
--- a/src/pages/api/requestContent.ts
+++ b/src/pages/api/requestContent.ts
@@ -1,7 +1,28 @@
 import type { NextApiRequest, NextApiResponse } from 'next';
 import axios from 'axios';
 
+import rateLimit from '../../utils/rate-limit';
+
+const limiter = rateLimit({
+    max: 600, // cache limit of 600 per 30 second period.
+    ttl: 30 * 1000,
+});
+
+const MAX_CONTENT_REQUEST_PER_PERIOD = 10;
+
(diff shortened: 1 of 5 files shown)


src/pages/api/requestContent.ts
```diff
@@ -1,7 +1,28 @@
...
+        await limiter.check(
+            res,
+            MAX_CONTENT_REQUEST_PER_PERIOD,
+            req.socket.remoteAddress || ''
+        );
+    } catch {
+        return res.status(500).json({
+            error: { message: 'Something went wrong' },
+        });
+    }
```
ajhenri, line 24: Wondering if this should be moved into a reusable helper function to be imported in all three cases

ajhenri (approved):

src/pages/api/requestContent.ts
mcolella14: So I was thinking about doing that, but I kinda wanted to leave it up to the endpoint to determine what the max-per-period requests are and what to do when that max has been reached. We might want to treat different endpoints differently in the future, so I figured I'd try to follow that pattern.

harikakotipalli (approved):

PR #202, opened 2022-06-15
Added structured logging in middleware

jackyw2018:
Added structured logging to NextJS middleware using the Pino library.


package.json
```diff
@@ -36,6 +36,7 @@
         "lru-cache": "7.10.1",
         "next": "12.1.0",
         "next-seo": "5.4.0",
+        "pino": "^8.0.0",
```
mcolella14, line 39: We are trying to keep all dependencies pinned to a specific version, so if we could do that with pino as well that would be great.

mcolella14 (approved):
Looks awesome! Thanks for doing this, adds a lot of value going forward. I just had some nitpicks, but overall LGTM. Also I noticed there is a merge conflict because I just merged something in the same file (sorry!), so lmk if you need any help with that.

src/utils/logger.ts
```diff
@@ -0,0 +1,60 @@
...
+        },
+    },
+    level: process.env.PINO_LOG_LEVEL || 'info',
+    // doc notes that formatting time in-process impacts logging performance
+    timestamp: stdTimeFunctions.isoTime, // built-in ISO 8601-formatted time in UTC
+});
+
+export default logger;
+
+// below is example usage of the logger
```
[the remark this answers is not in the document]
ajhenri: I think having one example is fine, but the rest of the comments could be replaced by a link to the documentation. Also, for multiline comments, use `/* ... */`.
```
    /* 
    * Example usage:
    * logger.info("hello world")
    * output: {"level":30,"time":2022-06-15T22:24:44.918Z,"msg":"hello world"}
    */
```

ajhenri (approved):
LGTM, aside from comments above

src/pages/_middleware.ts
```diff
@@ -1,5 +1,14 @@
...
 import { rewrites } from '../../config/rewrites';
+import logger from '../utils/logger';
+
+export function logDataFromNextReqAndRes(req: NextRequest, res: NextResponse) {
+    return {
+        url: req.nextUrl.pathname,
+        method: req.method,
+        statusCode: res.status,
+    };
+}
```
ajhenri, line 11: As you already pointed out in our meeting, this seems to be sort of a duplicate of `logDataFromNextAPIReqAndRes` that is in `logger.ts`. It seems the purpose of both  `logDataFromNextReqAndRes` and `logDataFromNextAPIReqAndRes` is the same but yet, there are different functions because of the different Next.js request types.

src/pages/api/updateFeedback.ts
```diff
@@ -18,15 +19,19 @@ export default async (req: NextApiRequest, res: NextApiResponse) => {
             req.socket.remoteAddress || ''
         );
     } catch {
-        return res.status(500).json({
+        res = res.status(500);
+        logger.error(logDataFromNextAPIReqAndRes(req, res));
```
ajhenri, line 23: With the way that this is invoked, you are 1) building a log output separately and 2) logging that output with the function here.

src/utils/logger.ts
```diff
@@ -0,0 +1,57 @@
...
+export function logDataFromNextAPIReqAndRes(
+    req: NextApiRequest,
+    res: NextApiResponse
+): ExtraLogData {
+    return {
+        url: req.url,
+        method: req.method,
+        statusCode: res.statusCode,
+    };
+}
```
ajhenri, line 33: The purpose of this module is to provide a logger, but here there is a log output builder specifically for request objects in Next.js, which seems a bit unrelated. What do you think about making this function log the output itself, rather than building a logging output?

As an example, I'm thinking somewhere along the lines of:
```
type RequestMethod = "GET" | "POST" | "PUT" | "PATCH";

export function logRequestData(
    url: string,
    method: RequestMethod,
    statusCode: number,
    logLevel: pino.Level,
): void {
    const logData = {
        url: url,
        method: method,
        statusCode: statusCode,
    };

    if(logLevel === 'error') logger.error(logData);
    else if(logLevel === 'info') logger.info(logData);
    else if(logLevel === 'warn') logger.warn(logData);
}
```

This way, the purpose of `logger.ts` is still only to provide utils specific to logging functionality, and we don't have to worry about different Next.js request objects. That logic doesn't need to be handled here.

jackyw2018: I agree with the change. It does separate the logic better and more cleanly. I can update the code based on this!

ajhenri (approved):
Nice work! Thanks for the prompt update. I like how you used a combination of both log level and status code, further improving upon the feedback provided!

PR #203, opened 2022-06-16
Add unit tests for components.

mcolella14:


harikakotipalli (approved):

ajhenri (approved):
This is awesome! Looks great.

harikakotipalli (approved):
Nicely done! one thought : when we have enough coverage  (90-95)+ we should make the check fail if it drops below 95%. That will enforce us to add enough testing.

mcolella14:
@harikakotipalli I agree, the coverage isn't great but I think once we start adjusting things and adding tests it should get better, and then we should definitely add the threshold.

PR #204, opened 2022-06-21
[DEVHUB-1323] Convert article page to use SSR

ajhenri:


ajhenri:
Only article w/ SSR with [l1_l2] routes:
Only article w/ SSR and removed [l1_l2] routes: https://devcenter-f799q3405-harikakotipalli.vercel.app/developer/products/mongodb/adl-sql-integration-test/

PR #205, opened 2022-06-22
[DEVHUB-1324] Replace babel with SWC

ajhenri:
Remove babel in favor of https://swc.rs/


mcolella14 (approved):
LGTM! Just the one question. Excited to see this in action!

PR #207, opened 2022-06-22
Jw/devhub 1294

jackyw2018:
To avoid /api/health/ flooding the logs, logic is added in the logger to skip logging /api/health/.


ajhenri (approved):
LGTM, aside from moving the `should_not_log` function to within `logRequestData` as discussed.

ajhenri (approved):
LGTM, thanks!

PR #208, opened 2022-06-23
Add headers to request logs

mcolella14:
I mainly want to be able to see if Kanopy passes a header with the client IP address. We don't have to log these headers after I find that out, but it might be nice to have that functionality just in case.


jackyw2018 (approved):
LGTM

PR #211, opened 2022-06-24
[DEVHUB-1286] Converting all pages to SSR

ajhenri:


src/page-templates/main-content-page/content-page-data.ts
```diff
@@ -0,0 +1,75 @@
...
+import { Crumb } from '../../components/breadcrumbs/types';
+import { appendDocumentationLinkToSideNav } from '../../utils/add-documentation-link-to-side-nav';
+import { determineVideoOrPodcast } from './content-page-template';
+import { getContentItemFromSlug } from '../../service/get-content-by-slug';
+
+let pluralize = require('pluralize');
+
+export const getContentPageData = async (slug: string[]) => {
+    const slugStr = slug.join('/');
+    const contentItem: ContentItem | null = await getContentItemFromSlug(
```
ajhenri, line 17: This will fetch only a single piece of content, which was previously using `getAllContentItems`.

src/page-templates/main-content-page/content-page-data.ts
```diff
@@ -0,0 +1,75 @@
...
+    setURLPathForNavItems(tertiaryNavItems);
+
+    const metaInfoForTopic = await getMetaInfoForTopic(sideNavFilterSlug);
+    tertiaryNavItems = appendDocumentationLinkToSideNav(
+        tertiaryNavItems,
+        metaInfoForTopic
+    );
+    const topicSlug = sideNavFilterSlug;
+    const topicName = metaInfoForTopic?.tagName ? metaInfoForTopic.tagName : '';
+    const relatedContent = getRelatedContent(
```
ajhenri, line 60: Since `getSideNav`, `getMetaInfoForTopic` and `getRelatedContent` are all pulling from the `.preval` data, these do not trigger Strapi API calls.

next.config.js
```diff
@@ -60,7 +62,7 @@ if (process.env.ANALYZE === 'true') {
     const withBundleAnalyzer = require('@next/bundle-analyzer')({
         enabled: true,
     });
-    module.exports = withBundleAnalyzer(configVals);
+    module.exports = withNextPluginPreval(withBundleAnalyzer(configVals));
 } else {
-    module.exports = configVals;
+    module.exports = withNextPluginPreval(configVals);
```
ajhenri, line 67: The preval data will be re-fetched at build time (every hour or so).

mcolella14 (approved):
Looks awesome! Thanks so much for doing this huge lift, super excited to see this in action!

next.config.js
harikakotipalli: what is preval data

ajhenri: It is static data that is used across multiple pages. Rather than requesting the data via Strapi each time, I am using this plugin that will prevent the additional API calls and just use static data retrieved during initial build. https://github.com/ricokahler/next-plugin-preval#why.

```
Site-wide data: if you have static data that's required across many different pages, getStaticProps is a somewhat awkward mechanism because for each new page, you'll have to re-fetch that same static data. For example, if you use getStaticProps to fetch content for your header, that data will be re-fetched on every page change.
```

harikakotipalli:
Thanks @ajhenri for doing this heavy lift. My only concern is the non reduction in the build time, even after reducing the number of static pages.

ajhenri:
> My only concern is the non reduction in the build time, even after reducing the number of static pages, other than that no issues.
Thanks, looking into it. It could be either related to the Apollo cache config updates or some other issue.

gfunk1230 (approved):
LGTM! thanks @ajhenri for taking on this.

harikakotipalli (approved):

src/pages/[...slug].tsx
```diff
@@ -1,128 +1,94 @@
...
-        <ContentPageTemplate
-            crumbs={crumbs}
-            topicSlug={topicSlug}
-            topicName={topicName}
-            contentItem={contentItem}
-            tertiaryNavItems={tertiaryNavItems}
-            relatedContent={relatedContent}
-        />
-    );
+    if (pageType === PageType.Topic) {
```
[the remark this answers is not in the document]
Comment (ajhenri) [reply src/pages/[...slug].tsx:21]: Great suggestion, agreed on using switch and abstracting this elsewhere! Will update

harikakotipalli (approved):

mcolella14 (approved):

ajhenri:
Thanks y'all! Going to merge this into staging

PR #212, opened 2022-06-27
Jw/devhub 1112

jackyw2018:
Per [ticket 1112](https://jira.mongodb.org/browse/DEVHUB-1112), Display date in each card is now updated date instead of published date, if updated date is available and more recent than the published date.


mcolella14 (approved):
I agree with Amanda's comments, but looks great to me!

ajhenri (approved):
Looks great. Thanks for making those changes!

PR #220, opened 2022-06-30
[DEVHUB-1320] Consume new featured endpoint.

mcolella14:
[DEVHUB-1320](https://jira.mongodb.org/browse/DEVHUB-1320)


src/service/get-all-featured.ts
```diff
@@ -1,16 +1,21 @@
...
     const featured = {
-        articles: featuredArticles ? featuredArticles : [],
-        podcasts: featuredPodcasts ? featuredPodcasts : [],
-        videos: featuredVideos ? featuredVideos : [],
+        articles: featuredArticles.map(({ title }) => title),
+        podcasts: featuredPodcasts.map(({ title }) => title),
+        videos: featuredVideos.map(({ title }) => title),
     };
     return featured;
 };
```
mcolella14, line 21: This isn't being used anymore, but I figured I'd leave it and its preval in there just in case we have a need for featured articles regardless of tag in the future.

gfunk1230 (approved):

PR #229, opened 2022-07-06
[DO NOT MERGE ]Devhub 1317

mcolella14:


gfunk1230 (approved):

ajhenri:
@mcolella14 Have you reviewed this with the Kanopy team?

mcolella14:
@ajhenri I have not, do you think it would be a good idea to? My assumption was that since this is within the confines of our staging namespace we didn't need to coordinate, but might be good to let them know? Also now that I'm thinking about it, I'm not even sure this will work in production, so might be good to have a conversation with them.

ajhenri:
@mcolella14 Yeah the main reason I ask this is because this endpoint will be publicly accessible over the internet. For this reason, I think it would be good to consult with Angelo or someone on the Kanopy team. They will also have more knowledge on best practices regarding using Drone for this and what other options we might have. In any case, we should get their go-ahead before moving forward with this.

src/pages/api/preview-build-webhook.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export default async (req: NextApiRequest, res: NextApiResponse) => {
+    if (req.method !== 'POST') {
+        return res.status(405).send('This is a POST-only endpoint.');
+    }
+
+    // If we don't have this header or if the event is not a PR somehow, don't trigger.
+    const event = req.headers['x-github-event'];
+    if (!event || Array.isArray(event) || event !== 'pull_request') {
+        return res.status(400).send('Event must be pull-requst');
```
redondos, line 57:
```suggestion
        return res.status(400).send('Event must be pull_request');
```

redondos (approved):

redondos:
> @mcolella14 Yeah the main reason I ask this is because this endpoint will be publicly accessible over the internet.

For additional security please file a KANOPY request to add an allowlist for this endpoint for Github's IP address space.

src/pages/api/preview-build-webhook.ts
```diff
@@ -0,0 +1,75 @@
...
+    if (headBranch !== 'DEVHUB-1317') {
+        return 'Did nothing because we are still testing this';
+    }
+
+    // Only build on opened, reopened, or synchronized.
+    const shouldBuild = ['opened', 'synchronize'].includes(body.action);
+
+    if (shouldBuild) {
+        // Trigger our custom event in our drone pieline.
+        const buildEndpoint = `https://drone.corp.mongodb.com/api/repos/mongodb/devcenter/builds?branch=${headBranch}&number=${body.number}`;
```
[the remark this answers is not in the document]
mcolella14: Yes there is verification as specified by the webhook payload docs for pull_request events. `number` is specified as an integer https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#pull_request

mcolella14:
Leaving this a draft so it doesn't accidentally get merged, but I believe all we're waiting on is https://jira.mongodb.org/browse/KANOPY-160, then we can merge and swap the webhook url to the production Kanopy deployment.

.drone.yml
```diff
@@ -330,3 +330,93 @@ trigger:
...
+      NPM_AUTH:
+        from_secret: NPM_AUTH
+      NPM_EMAIL:
+        from_secret: NPM_EMAIL
+      REALM_API_URL:
+        from_secret: REALM_API_URL
+      REALM_SEARCH_URL:
+        from_secret: REALM_SEARCH_URL
+      STRAPI_URL:
+        from_secret: STRAPI_URL_PROD
```
ajhenri, line 358: Just a heads up, I recently merged changes to update this to use `STRAPI_URL_STAGING` (https://github.com/mongodb/devcenter/pull/238). I think for previews we want this as well, since we now have a way to refresh staging DB.

ajhenri (changes requested):

ajhenri (approved):

PR #236, opened 2022-07-11
[DEVHUB-1314] Add Prometheus monitoring to devcenter Next.js app

ajhenri:
Reuse metrics endpoint that was created in another ticket by Jacky. Defines configuration rules for metrics monitoring and sets up basic Alertmanager rules for service uptime and container readiness.

Update staging CDN link based on Kanopy advisory.

Example query:
`up{namespace="devrel", job="devcenter-frontend-ah-web-app"}`


src/pages/api/metrics.ts
```diff
@@ -0,0 +1,25 @@
+import { NextApiRequest, NextApiResponse } from 'next';
+import { Registry } from 'prom-client';
+import { prometheus } from '../../config/prometheus';
+
+const metricsApi = async (req: NextApiRequest, res: NextApiResponse) => {
+    const authHeader = req.headers.authorization;
+    if (!authHeader) return res.status(400).send('Bad Request');
```
mcolella14, line 7: This is incredibly nitpicky, but I think it should technically be a 401 too since it's an unauthorized request.

mcolella14 (approved):
Looks great! Will feel better knowing we have some alerts now.

gfunk1230 (approved):

src/pages/api/metrics.ts
ajhenri: Not nitpicky at all, thanks for spotting!

jackyw2018 (approved):
LGTM

PR #248, opened 2022-07-19
[DEVHUB-1356] Implement and Configure Sentry IO into Developer Center

ajhenri:
- Improve drone pipeline by removing need for build args (student app does something similar, so most of the Drone pipeline code is taken from there).
- Adds Sentry monitoring to the project: includes ErrorBoundary, hooks on API handlers and handling on all error pages.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,0 @@
-node_modules/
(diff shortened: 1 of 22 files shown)


mcolella14 (approved):
LGTM!

PR #251, opened 2022-07-21
[DEVHUB-1340] Add Sort-By Feature

jackyw2018:
Per the [ticket](https://jira.mongodb.org/browse/DEVHUB-1340?jql=assignee%20in%20(currentUser())), the sort-by front-end is added to both search page and topic content type page.

To test, please change the search url in env:
- REALM_SEARCH_URL=https://us-east-1.aws.data.mongodb-api.com/app/devhub-search-service-stage-rqnmi/endpoint

Given that very few contents have ratings currently, to test between highest rated and most recent:
- Please filter for javascript without search query


src/components/search/utils.ts
```diff
@@ -0,0 +1,5 @@
+export const sortByOptions: { [key: string]: number } = {
```
ajhenri, line 1: Would it be accurate to say this is a `{ [key: SortByType]: number }`?

src/hooks/search/index.ts
```diff
@@ -62,6 +68,21 @@ const useSearch = (
         }
     };
 
+    const onSort = (sortByValue: string) => {
+        setResultsToShow(10);
+        setSortBy(sortByValue as SortByType);
+        console.log(sortByValue);
```
ajhenri, line 74: Nit: console.log

ajhenri (commented):
Great start! I am unable to see the filter on the search page, is it just me? Do you have screenshots of the way this looks on desktop vs mobile?

Also, I realize, we did not deploy the search backend changes yet to prod, as they are only in staging right now. We'll need to do this before merging this PR.

src/components/search/search.tsx
```diff
@@ -68,15 +79,36 @@ const Search: React.FunctionComponent<SearchProps> = ({
...
+                </div>
+                <Select
+                    sx={sortBoxStyles}
+                    label="Sort by"
+                    name="sort-by-dropdown"
+                    options={Object.keys(sortByOptions)}
+                    value={sortBy}
+                    onSelect={onSort}
+                    width="100%"
+                    height="100%"
```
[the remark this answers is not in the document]
jackyw2018: The current version is shown by the screenshot which should follow the [Figma design](https://www.figma.com/file/55scZezlUzn2O7VDvQzeWU/DevCenter-V1?node-id=3101%3A843569).

image

jackyw2018:
> Great start! I am unable to see the filter on the search page, is it just me? Do you have screenshots of the way this looks on desktop vs mobile?
> 
> Also, I realize, we did not deploy the search backend changes yet to prod, as they are only in staging right now. We'll need to do this before merging this PR.

Yes, but I do note that the app should not fail if we merge frontend changes first. It will send a request with an additional/new query param (sortMode) to the existing realm function (without the backend change), which will simply ignore the additional param. 

This does however mean sort-by feature on frontend will not change content order when clicked.
@ajhenri

.drone.yml
```diff
@@ -20,7 +20,7 @@ steps:
       REALM_API_URL:
         from_secret: REALM_API_URL
       REALM_SEARCH_URL:
-        from_secret: REALM_SEARCH_URL
```
mcolella14, line 23: Just curious, is the intention to update the existing realm search function with the contents on the staging one? If so, just want to be sure that we set this back.

src/components/search/styles.ts
```diff
@@ -9,16 +9,24 @@ export const titleStyles = {
     gap: 'inc30',
 };
 
+export const searchBoxSortBarWrapperStyles = {
+    display: 'grid',
+    gridTemplateColumns: '',
```
mcolella14, line 14: I think this can be removed.

mcolella14 (approved):
Looks great to me!

.drone.yml
ajhenri: @mcolella14 I asked to add this since we don't have a staging version.

ajhenri:
Looks good to me. Will work with you tomorrow on deployment process for this.

ajhenri (approved):
Good to merge now! Nice work!

PR #259, opened 2022-07-29
[CHORE] Upgrade flora, account for link styling

mcolella14:
Per Angelo's suggestion, I went through all uses of the Flora Link component and made any changes to account for the breaking change they introduced in v0.22.5. They reverted the change in their most recent release (0.22.6), so it doesn't affect anything in this PR, but we will be ready when they reimplement that change.


ajhenri (approved):
LGTM, thanks for working on this!

PR #266, opened 2022-08-08
Fix ERR_UNHANDLED_REJECTION

ajhenri:
Often, when `yarn build` and tests are run, the build RSS feed script fails on the following error "AxiosError: maxContentLength size of Infinity exceeded" (examples: https://github.com/mongodb/devcenter/runs/7726957248?check_suite_focus=true, https://github.com/mongodb/devcenter/runs/7581091449?check_suite_focus=true, )
```
You can learn more, including how to opt-out if you'd not like to participate in this anonymous program, by visiting the following URL:
https://nextjs.org/telemetry

node:internal/process/promises:279
            triggerUncaughtException(err, true /* fromPromise */);
            ^

[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "AxiosError: maxContentLength size of Infinity exceeded".] {
  code: 'ERR_UNHANDLED_REJECTION'
}
```

This changes the part of my last PR, which was attempted in order to get tests to pass, but doesn't seem to work to resolve the above issue. 
It seems this is an issue that is unresolved in axios (issue 4806), so I am removing it's use and replacing with the built in node.js https module.

### DIFF
--- a/src/scripts/build-rss-feed.js
+++ b/src/scripts/build-rss-feed.js
@@ -18,6 +18,9 @@ async function buildRssFeed(baseUrl) {
         },
     });
 
+    // We don't use axios here due to a bug in axios where
+    // it cannot always handle responses with long content lengths.
+    // Issue referenced in https://github.com/axios/axios/issues/4806.
     https
         .get(
             `${process.env.REALM_SEARCH_URL}/search_devcenter?s=`,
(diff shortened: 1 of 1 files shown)


mcolella14 (approved):
LGTM, I'd say maybe just leave a comment explaining the scenario, because it could be confusing just looking at this why we're not using axios.

PR #217, opened 2022-06-29
Devhub 1041 sso

harikakotipalli:

PR #271, opened 2022-08-10
[DEVHUB-1041] SSO for DevCenter

ajhenri:
Implements SSO (Single Sign On) for DevCenter.

- Includes installation and setup of next-auth and leafygreen-ui libary for the UserMenu component (POC in https://github.com/mongodb/devcenter/pull/217).
- Created a signin page that will perform the signin process on the client side (this is because next-auth doesn't provide capability to signin from the server)
- Added leafygreen menu to desktop and mobile views.
- Integrate next-auth with middleware and set up session profile values.
- Keeps track of the previously visited page when signing in

Preview Environment: https://devcenter-frontend-sso.devrel.staging.corp.mongodb.com/developer/
CDN: https://mongodbcom-cdn.staging.corp.mongodb.com/developer/


src/components/layout.tsx
```diff
@@ -1,18 +1,54 @@
...
+
     return (
         <>
             <Global styles={globalStyles} />
             <UnifiedNavCustom
                 position="sticky"
                 floraTheme="default"
                 property={{ name: 'DEVHUB', searchParams: [] }}
+                hideTryFree={!!session}
+                hideSignIn={!!session}
```
ajhenri, line 50: There is currently a FOUC due to some of the pages still being statically rendered for page speed and optimization. As part of [DEVHUB-1290](https://jira.mongodb.org/browse/DEVHUB-1290) I am going to be using `next-plugin-preval` to make all static pages server-side rendered-- this is because we still need to optimize a lot of the queries on the pages such as topic and topic content (if we make these SSR, they will be very slow). However if we use a combination of preval with SSR, we can avoid this issue while still rendering the session server-side.

src/components/seconardynavnew/mobile.tsx
```diff
@@ -331,17 +342,19 @@ const MobileView = () => {
...
                     </TypographyScale>
                     {!mobileMenuIsOpen && (
-                        <SystemIcon
-                            sx={chevronStylesForMainLink}
-                            className="chevron-icon"
-                            name={ESystemIconNames.CHEVRON_DOWN}
-                            size="small"
-                            strokeWeight="large"
-                        />
+                        <>
```
justinmckenzie, line 345: you might be able to skip the fragment if there's just one child element

src/components/seconardynavnew/nav-item.tsx
```diff
@@ -10,8 +11,15 @@ const StyledList = {
     minHeight: [null, null, null, null, '37px'],
 };
 
-const SecondaryLinksList: React.FunctionComponent<IProps> = ({ children }) => {
-    return <li sx={StyledList}>{children}</li>;
+const SecondaryLinksList: React.FunctionComponent<IProps> = ({
+    children,
+    linkClassName,
+}) => {
```
justinmckenzie, line 17: you can do implicit return here, up to you

src/pages/_middleware.ts
```diff
@@ -112,5 +113,24 @@ export async function middleware(req: NextRequest) {
 
     const res = NextResponse.next();
     logRequestData(pathname, req.method, res.status);
+
+    // @ts-ignore
+    if (req.nextauth && req.nextauth.token) {
```
justinmckenzie, line 118: you might be able to do option chaining here  `if (req?.nextauth?.token) { ... }`

src/pages/auth/signin.tsx
```diff
@@ -0,0 +1,72 @@
...
+    // The sign in has to be done client-side since next-auth does not
+    // support a way to log in server side. This page will "render"
+    // the login process and then redirect the client to either the
+    // account portal or the previously visited page (if already logged in).
+    useEffect(() => {
+        if (status === 'unauthenticated') {
+            const callbackUrl = getURLPath(fromPagePath);
+            // https://github.com/nextauthjs/next-auth/issues/45
+            // Note: next-auth currently has no way of doing a server side signIn()
+            signIn('okta', { callbackUrl: callbackUrl });
```
justinmckenzie, line 34: you can do shorthand here `signIn('okta', { callbackUrl: callbackUrl });`

justinmckenzie (approved):
LGTM! Only comments were optional code syntax stuff from me

mcolella14 (approved):
LGTM! Awesome work, excited to see this in action!

src/pages/_middleware.ts
ajhenri: Thanks, great point! I am actually going to delete this block since I will not be using the cookie.

src/components/seconardynavnew/mobile.tsx
ajhenri: Nice catch!

PR #275, opened 2022-08-17
[DEVHUB-1290] Make all content crawlable on search pages

ajhenri:
This contains changes to make all content crawl-able on all search pages. I split the refactoring of search front-end and server side pagination to a different ticket (https://jira.mongodb.org/browse/DEVHUB-1482) that I am working on at the same time (https://github.com/mongodb/devcenter/compare/ah/devhub-1290...ah/devhub-1290-search-refactor). Currently having issues with optimizing `useSWRInfinite` with server side pagination, so I split this work away of the SEO requirements here. 

The changes in this PR reflect only what is needed for SEO, including:

- Making searchable content pages SSR rendered
- Setting up "initial" data and handling fetch failures (possible to refactor this with https://swr.vercel.app/docs/with-nextjs#pre-rendering-with-default-data, which will be done in DEVHUB-1482, but we still need the fetch functionality added here)
- Setting page number in query string working with Load More anchor
- Setting page title and canonical URLs based on page number

What's not included:
- Filtering and sorting is used among the search pages, but these are not required to be crawled for SEO, so only paging on the URIs/canonicals/titles is necessary.
- Rewriting of the search backend (remove existing sorting implementation and add pagination) (This will be included in above refactor ticket https://github.com/10gen/devhub-api-service/compare/ah/devhub-1290)
- Replacing useSWR with useSWRInfinite (I am having issues with search optimization using this, so I need more time to fix before moving forward with this refactor).

This essentially is a hybrid of SSR and client side rendering behavior for search, and this seems optimal as we do not want page reloads on each filter / page / sort.


src/components/search/search.tsx
```diff
@@ -59,10 +127,56 @@ const Search: React.FunctionComponent<SearchProps> = ({
...
+                    pathname: router.pathname,
+                    query: query,
+                },
+                undefined,
+                {
+                    scroll: false,
+                    shallow: true,
+                }
+            );
+            setResultsToShow(currentPage * DEFAULT_PAGE_SIZE + 10);
```
ajhenri, line 161: After refactoring, this will be replaced with `setSize` from useSWRInfinite. The router handling and page number tracking will still be needed though for SEO.

src/api-requests/get-all-search-content.ts
```diff
@@ -1,12 +1,46 @@
-import axios from 'axios';
+import {
+    buildSearchQuery,
+    SearchQueryParams,
+} from '../components/search/utils';
 import { SearchItem } from '../components/search/types';
 
 // Should find a way to cache this response. Only use it in a few places for search filters but still.
 export const getAllSearchContent = async (): Promise<SearchItem[]> => {
```
justinmckenzie, line 8: `getAllSearchContent` and `getSearchContent` are two very similar functions. Could we possibly look into combining the two that would just take in that `queryParams` arg?

src/api-requests/get-all-search-content.ts
```diff
@@ -1,12 +1,46 @@
...
+    const url = `${process.env.REALM_SEARCH_URL}/search_devcenter?s=`;
+    const options = {
+        method: 'GET',
+        headers: {
+            Accept: 'application/json',
+            'Content-Type': 'application/json',
+        },
+    };
+
+    return fetch(url, options)
```
justinmckenzie, line 18: if we're moving away from axios, we might want to add a safeguard here to handle failures, and we can also utilize `async/await` since it's already declared on the function

```
try {
   const req = await fetch(url, options);
   // do if (req.status === 200 || req.ok) check if you want
   const { data } = await req.json();
   return data;
} catch (e) { ... }
```

src/components/search/search.tsx
```diff
@@ -47,9 +70,54 @@ const Search: React.FunctionComponent<SearchProps> = ({
...
-    } = useSearch(contentType, tagSlug);
+    } = useSearch(
+        contentType,
+        tagSlug,
+        undefined,
+        initialSearchData ? pageNumber : undefined
+    );
+
+    const hasEmptyFilterAndQuery = () => {
+        return (!searchString || searchString == '') && allFilters.length == 0;
```
justinmckenzie, line 81: use `===` unless there was a specific reason to not use it

src/components/search/search.tsx
```diff
@@ -141,31 +261,37 @@ const Search: React.FunctionComponent<SearchProps> = ({
                 </div>
             )}
             <div sx={{}}></div>
-            {!!numberOfResults || isValidating || error ? (
+            {!!getResultData().length || getResultIsValidating() || error ? (
```
justinmckenzie, line 264: should we move this before the return and set it in a variable, that way we are not calling the function multiple times (also on line 267, line 280)

src/hooks/search/index.ts
```diff
@@ -93,126 +99,132 @@ const useSearch = (
...
+            product,
+            language,
+            technology,
+            contributedBy,
+            contentType,
+            expertiseLevel,
+        } = router.query;
+
+        let allNewFilters: FilterItem[] = [];
+        if (product) {
```
justinmckenzie, line 121: might be a little bit of a pain to do, but this `if` and the `if` on line 147 look pretty similar (unless im missing something), they could possibly be extracted to a function and then just passed the relevant arrays, etc. 

Same with the if statements from 174 through 212

src/hooks/search/utils.ts
```diff
@@ -52,10 +69,14 @@ const searchItemToContentItem = ({
...
 
-export const getFilters = async (contentType?: PillCategory) => {
+export const getFilters = async (
+    contentType?: PillCategory,
+    allSearchContent?: SearchItem[]
+) => {
     const allFilters = contentType === undefined;
-    const allContent: SearchItem[] = await getAllSearchContent();
-
+    const allContent: SearchItem[] = allSearchContent
```
justinmckenzie, line 77: `allSearchContent || await getAllSearchContent()`

src/page-templates/content-type/content-type.tsx
```diff
@@ -75,7 +102,28 @@ const ContentTypePage: NextPage<ContentTypePageProps> = ({
...
+        return (!searchString || searchString == '') && allFilters.length == 0;
+    };
+
+    const getResultData = () => {
+        return initialSearchData && hasEmptyFilterAndQuery()
+            ? initialSearchData
+            : data;
+    };
+
+    const getResultIsValidating = () => {
```
justinmckenzie, line 122: if there's project-wide util files available, these can be combined with the functions in `search.tsx`

src/page-templates/content-type/content-type.tsx
```diff
@@ -301,33 +423,41 @@ const ContentTypePage: NextPage<ContentTypePageProps> = ({
...
-                                                    setResultsToShow(
-                                                        resultsToShow + 10
-                                                    )
-                                                }
-                                                variant="secondary"
-                                            >
-                                                Load more
-                                            </Button>
-                                        )}
+                                        {!getResultIsValidating() &&
```
justinmckenzie, line 443: same as above, we can store these in variables to avoid multiple calls

src/api-requests/get-all-search-content.ts
ajhenri: Agreed, will fix that.

src/components/search/search.tsx
ajhenri: Good call, I think I had it that way and can't recall why I changed this to a function. Will update.

src/hooks/search/utils.ts
ajhenri: Good catch on this one

src/page-templates/content-type/content-type.tsx
ajhenri: @justinmckenzie These functions are repeated but I don't think there will be much gain to moving them to a utils because the dependencies are specific to each component (e.g. isValidating, initialSearchContent, data, searchString, allFilters) and each component is technically repeated but slightly different, so I would need to pass 5 parameters and we'd only be repeating ternary operator logic. I did however, move `hasEmptyFilterAndQuery` to the hook utilities file.

ajhenri: Ended up moving these out to utils anyway, but added a note for this to be refactored at a later time due to the amount of parameters. I think this might actually be removed in the refactor PR mentioned above.

justinmckenzie: @ajhenri sounds good...i was just making the suggestion because I noticed the repetition, if it doesn't really make sense to do it I'm totally cool with whatever approach you think is best.

justinmckenzie (approved):

src/page-templates/content-type/content-type.tsx
ajhenri: Haha sure yeah, it was already done. Does look much better without the repetition so it was a great suggestion

src/components/search/search.tsx
```diff
@@ -141,31 +268,40 @@ const Search: React.FunctionComponent<SearchProps> = ({
...
+                                    href={
+                                        hasEmptyFilterAndQuery(
+                                            searchString,
+                                            allFilters
+                                        )
+                                            ? `/developer${path}/?page=${
+                                                  currentPage + 1
+                                              }`
+                                            : '#'
                                     }
```
mcolella14, line 298: nitpick, but might be better to calculate this value above the render method to keep the DOM structure easy to read.

src/hooks/search/index.ts
```diff
@@ -93,126 +99,122 @@ const useSearch = (
         };
     }, []);
 
+    const getFiltersFromQueryStr = () => {
```
mcolella14, line 102: Thanks for refactoring this, looks so much cleaner in the `useEffect` now.

src/hooks/search/utils.ts
```diff
@@ -218,6 +238,43 @@ const itemInFilterGroup = (tags: Tag[], filters: FilterItem[]) => {
...
+                  ? (initialPageNumber - 1) * DEFAULT_PAGE_SIZE
+                  : 0
+          );
+};
+
+export const getResultIsValidating = (
+    initialSearchData: ContentItem[] | undefined,
+    searchString: string,
+    allFilters: FilterItem[],
+    earchString: string,
```
mcolella14, line 270: Looks like `earchString` was accidentally left in here

mcolella14 (approved):
LGTM, just a couple nitpicks. Awesome work with this, seems like it was a lot so thank you and sorry for the initial spaghetti code lol

src/hooks/search/utils.ts
ajhenri: Oh wow, haha. Somehow all of this was still working properly too.

ajhenri:
@justinmckenzie @mcolella14 Thanks for the reviews and great feedback, I know it was a ton of code!

PR #279, opened 2022-08-18
[DEVHUB-1167] - Make secondary nav sticky instead of consistent nav, lock background scrolling on overlays.

mcolella14:
[DEVHUB-1167](https://jira.mongodb.org/browse/DEVHUB-1167)

The first commit was just making the secondary nav sticky over the consistent nav. 

The second commit centers around an issue I noticed where on the mobile nav menu (as well as on the mobile search filter overlay), if a user scrolls to the bottom of the overlay, they can keep scrolling the page in the background. This is just a weird experience, so I added a context value so we have some global knowledge of whether an overlay is present. We then lock the main body from scrolling if this is the case.


src/components/seconardynavnew/mobile-styles.ts
```diff
@@ -1,11 +1,23 @@
 import { ThemeUIStyleObject } from 'theme-ui';
 
+export const navWrapperStyles: ThemeUIStyleObject = {
+    boxSizing: 'border-box',
+    background: '#fff',
+    backgroundColor: 'white',
```
justinmckenzie, line 6: why `background` and `backgroundColor`?

src/components/seconardynavnew/mobile-styles.ts
```diff
@@ -1,11 +1,23 @@
...
+    display: ['block', 'block', 'block', 'none'],
```
justinmckenzie, line 7: `display: ['block', null, null, 'none']` (unless im misunderstanding that on my end)

justinmckenzie (approved):
Nitpicks from existing CSS, everything else looks great

src/components/seconardynavnew/mobile-styles.ts
mcolella14: no idea lol, I'll correct this

src/components/seconardynavnew/mobile-styles.ts
mcolella14: ah yep good catch!

gfunk1230 (approved):

justinmckenzie (approved):

PR #281, opened 2022-08-24
[DEVHUB-1318]: Improve and Correct the Code Examples Experience

justinmckenzie:
Ticket: https://jira.mongodb.org/browse/DEVHUB-1318

Scope
- Adds a 2nd group of CTA's for Code Example content types to view code on github or try out live site.
- Moves tags to aside view on desktop with labeled grouping -> falls back to original placement of tags below Author Info
- Adds unit test for rendering with labels to `tag-section`

### DIFF
--- a/src/page-templates/main-content-page/content-page-template.tsx
+++ b/src/page-templates/main-content-page/content-page-template.tsx
@@ -393,7 +393,7 @@ const ContentPageTemplate: NextPage<ContentPageProps> = ({
                     linkIcon="arrow"
                     href={getCtaLinkForVideosOrPodcasts(category)}
                     sx={{
-                        display: 'inline-block',
+                        display: 'block',
                         marginTop: 'inc40',
                     }}
                 >
--- a/src/page-templates/main-content-page/content-page-template.tsx
+++ b/src/page-templates/main-content-page/content-page-template.tsx
@@ -40,7 +40,6 @@ import { getPlaceHolderImage } from '../../utils/get-place-holder-thumbnail';
 import Image from 'next/image';
 import { getCardProps, thumbnailLoader } from '../../components/card/utils';
 import parse from 'html-react-parser';
-import CTALink from '../../components/hero/CTALink';
 import { DocumentBody } from '../../components/article-body/document-body';
 import SeriesCard from '../../components/series-card';
 import { Grid, ThemeUICSSObject } from 'theme-ui';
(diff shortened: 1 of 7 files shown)


src/page-templates/main-content-page/content-page-template.tsx
```diff
@@ -224,6 +227,44 @@ const ContentPageTemplate: NextPage<ContentPageProps> = ({
...
+                    iconName={ESystemIconNames.EXTERNAL}
+                    sx={{
+                        width: 'auto',
+                    }}
+                >
+                    View Code
+                </Button>
+            )}
+            {liveSiteUrl && (
+                <Link
```
justinmckenzie, line 255: @mcolella14 previous implementation was originally using the `<CTALink />` element here, which looks exactly like the Flora `<Link linkIcon="arrow"` so I used that here. Was wondering the reasoning for creating the custom `<CTALink />` component? Looks like it's used only in one or two other places in the codebase.

mcolella14: Yeah I think we originally had some more styling or functionality on it, but then removed it and didn't refactor, so I'm fine with just taking that out.

src/page-templates/main-content-page/content-page-template.tsx
```diff
@@ -314,6 +361,11 @@ const ContentPageTemplate: NextPage<ContentPageProps> = ({
                     )}
                 </div>
                 {!previewMode && ratingSection}
+                {(githubUrl || liveSiteUrl) &&
```
mcolella14, line 364: Might be a good idea to throw in an `isCodeExample` condition on this and line 401, just to be super safe in case a `githubUrl` or `liveSiteUrl` accidentally gets added to a non-Code Example piece.

mcolella14 (approved):
Looks awesome!

justinmckenzie:
@mcolella14 replaced `<CTALink />` with Flora `<Link />` so just want you to gloss over the newest commit just to make sure everything looks good

mcolella14 (changes requested):
There are actually some changes with CTALink we would need to account for, check out this article and the CTA at the bottom https://www.mongodb.com/developer/videos/serverless-functions-explained---mongodb-realm-backend-as-a-service---jumpstart/

mcolella14 (approved):

PR #284, opened 2022-08-26
[DEVHUB-1431] Fix having to click twice for videos to play

max-melo:
[[DEVHUB-1431] Must click "play" twice for a video to play](https://jira.mongodb.org/browse/DEVHUB-1431)

Before, there was a thumbnail that had to be clicked on videos:
Screen Shot 2022-08-26 at 10 11 15 AM

Now, it displays the video directly from the provider:
Screen Shot 2022-08-26 at 10 10 27 AM

### DIFF
--- a/src/components/article-body/body-components/video-embed.tsx
+++ b/src/components/article-body/body-components/video-embed.tsx
@@ -36,7 +36,7 @@ export const VideoEmbed = ({
 }: {
     argument: Argument[];
     name: string;
-    thumbnail: string;
+    thumbnail?: string;
 }) => {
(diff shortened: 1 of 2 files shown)


justinmckenzie (approved):
This is fine with me, I'm just trying to think who we should run this by from the biz side to make sure its okay since I think this ticket was logged by someone from Dev Rel

mcolella14 (commented):
I believe the designs called for a specific thumbnail overlay image to be present, so I would definitely check with mike before merging this.

max-melo:
Checked with mike, he said the change is good to go

PR #291, opened 2022-09-01
[DEVHUB-1247] Set custom cookie names in next-auth config

ajhenri:
Overrides cookie names for next-auth.pkce.code_verifier and next-auth.state with custom set names. If secure cookies are enabled, __Session is prepended.

This was needed to get the staging CDN to work properly given the existing key list Kanopy configured.

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -34,7 +34,7 @@ const configVals = {
                 headers: [
                     {
                         key: 'Cache-Control',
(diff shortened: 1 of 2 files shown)


ajhenri:
Decided to just add __Secure-next-auth.pkce.code_verifier and __Secure-next-auth.state to the list instead.

PR #292, opened 2022-09-01
[DEVHUB-1125] Add Sitemap

max-melo:
[[DEVHUB-1125] NextJS App - Add sitemap](https://jira.mongodb.org/browse/DEVHUB-1125)

Add a sitemap to the site which will reside at [https://mongodb.com/developer/sitemap/sitemap-index.xml](https://mongodb.com/developer/sitemap/sitemap-index.xml) as per the sitemap entry in MongoDB's [main sitemap index file.](https://www.mongodb.com/sitemap.xml)

`next-sitemap` has the ability to generate static sitemaps via a config file, but I didn't use that because of the need for dynamic routes. Their documentation states that in order to have a static and dynamic sitemap in the same app, they must live in separate sitemap files and be referenced by `robots.txt` ([which isn't controlled by this app](https://www.mongodb.com/robots.txt)) and/or referenced by an index sitemap (which also isn't possible, because [index sitemaps shouldn't be nested](https://support.google.com/webmasters/answer/7451001#errors&zippy=%2Cerror-list)).  Because of this, the only solution I found is to manually generate all of the paths dynamically, using `fs` and some of the service functions.

I added priority values to match what seems like the hierarchy of the pages. Main sections of the site, being the most important, I gave priority `1`, specific topics being less important were given `0.7` (`next-sitemap`'s default value), and individual content pages were given `0.5` (the default value in the spec). I can recalibrate these if needed, or remove them altogether as the field is optional.

[Here's a sample of what is generated](https://pastebin.com/raw/xE9ZKi2K).

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
         "next-auth": "4.10.3",
         "next-plugin-preval": "1.2.4",
         "next-seo": "5.4.0",
-        "next-sitemap": "^3.1.21",
+        "next-sitemap": "3.1.21",
         "pino": "8.0.0",
         "pluralize": "8.0.0",
         "prom-client": "14.0.1",
--- a/package.json
+++ b/package.json
@@ -4,7 +4,6 @@
     "scripts": {
         "dev": "next dev",
         "build": "next build",
-        "postbuild": "next-sitemap --config next-sitemap.config.js",
         "build:analyze": "ANALYZE=true next build",
         "start": "next start",
         "lint": "next lint",
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
     "scripts": {
         "dev": "next dev",
         "build": "next build",
+        "postbuild": "next-sitemap --config next-sitemap.config.js",
         "build:analyze": "ANALYZE=true next build",
         "start": "next start",
         "lint": "next lint",
@@ -41,6 +42,7 @@
         "next-auth": "4.10.3",
         "next-plugin-preval": "1.2.4",
         "next-seo": "5.4.0",
+        "next-sitemap": "^3.1.21",
         "pino": "8.0.0",
         "pluralize": "8.0.0",
         "prom-client": "14.0.1",
--- a/src/pages/sitemap/sitemap-index.xml.ts
+++ b/src/pages/sitemap/sitemap-index.xml.ts
@@ -2,18 +2,16 @@ import fs from 'fs';
 import path from 'path';
 import { getServerSideSitemap } from 'next-sitemap';
 import { GetServerSideProps } from 'next';
-import { getAllContentItems } from '../../service/get-all-content';
+import allContentPreval from '../../service/get-all-content.preval';
 import { getTopicPagePathMappings } from '../../service/get-topic-paths';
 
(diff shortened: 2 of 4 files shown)


justinmckenzie (approved):

src/pages/sitemap/sitemap-index.xml.ts
```diff
@@ -0,0 +1,55 @@
+import fs from 'fs';
+import path from 'path';
+import { getServerSideSitemap } from 'next-sitemap';
+import { GetServerSideProps } from 'next';
+import { getAllContentItems } from '../../service/get-all-content';
+import { getTopicPagePathMappings } from '../../service/get-topic-paths';
+
+const DEVCENTER_URL =
+    process.env.DEVCENTER_URL || 'https://mongodb.com/developer';
```
ajhenri, line 9: Where is `process.env.DEVCENTER_URL` set? Do you mean `HOST_URL`?

src/pages/sitemap/sitemap-index.xml.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export const getServerSideProps: GetServerSideProps = async context => {
+    const curDate = new Date().toISOString();
+
+    const { topicContentTypePaths, topicPaths } =
+        await getTopicPagePathMappings();
+    const allContent = await getAllContentItems();
```
ajhenri, line 16: How fresh does the sitemap need to be? `getAllContentItems` will make a heavy query, but we have that data pre-evaluated (example usage here: https://github.com/mongodb/devcenter/blob/f7d9407216e9d1f26ee477eff6b4894665d9ac1e/src/page-templates/topic-page/topic-page-data.ts#L10) so that in a way, it is static data "cached" in Next.js for future usage.

max-melo: wasn't aware of this, nice!

src/pages/sitemap/sitemap-index.xml.ts
max-melo: I didn't use `HOST_URL` because I wanted the sitemap to have the correct values locally. my understanding is that locally & in staging we'd want to see the production links and not the ones specific to that environment (e.g. `https://mongodb.com/developer/articles` instead of `localhost:3000/articles`)/

`DEVCENTER_URL` is just an unused env variable name that I came up with for an extra bit of configurability, but it probably isn't necessary as I'm assuming the URL to the site likely won't change, and if it does it can be changed in the code. I'm happy to remove this.

package.json
```diff
@@ -41,6 +41,7 @@
         "next-auth": "4.10.3",
         "next-plugin-preval": "1.2.4",
         "next-seo": "5.4.0",
+        "next-sitemap": "^3.1.21",
```
mcolella14, line 44: We try to pin all of our non-dev dependencies, would be nice to have this follow suit.

mcolella14 (commented):
Just the pinning and the stuff Amanda said, otherwise looks great!

ajhenri (approved):
LGTM

mcolella14 (approved):

PR #297, opened 2022-09-08
[DEVHUB-1482] Replace search with search v2

ajhenri:
## Jira Ticket:

Part of DEVHUB-1482. Associated backend changes are in https://github.com/10gen/devhub-api-service/pull/29.

## Description:

This replaces the search endpoint to use POST method. It will be needed for the future changes that involve refactoring the search to useSWRInfinite.

The reason for using POST is so that we can pass the filters (of which can be many, as much as ~30) into the POST body rather than as a query string (which would be too long).

-   [ ] I have added tests that prove my fix is effective or that my feature works - Tests already exist for search, so here I am just swapping out the endpoints. It should behave the same as earlier.

### DIFF
--- a/src/hooks/search/utils.ts
+++ b/src/hooks/search/utils.ts
@@ -311,9 +311,12 @@ export const itemInFilters = (
 };
 
 export const fetcher: Fetcher<ContentItem[], string> = queryString => {
-    return fetch(
-        (getURLPath('/api/search') as string) + '?' + queryString
-    ).then(async response => {
+    return fetch((getURLPath('/api/search') as string) + '?' + queryString, {
+        method: 'POST',
+        body: JSON.stringify({
+            filters: [],
+        }),
+    }).then(async response => {
         const r_json: SearchItem[] = await response.json();
         return r_json.map(searchItemToContentItem);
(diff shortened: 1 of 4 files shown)


src/hooks/search/utils.ts
```diff
@@ -311,9 +311,12 @@ export const itemInFilters = (
 };
 
 export const fetcher: Fetcher<ContentItem[], string> = queryString => {
-    return fetch(
-        (getURLPath('/api/search') as string) + '?' + queryString
-    ).then(async response => {
+    return fetch((getURLPath('/api/search') as string) + '?' + queryString, {
```
justinmckenzie, line 314: super nitpick but you could use string template here instead of concatenation

justinmckenzie (approved):

src/hooks/search/utils.ts
ajhenri: Yeah, it definitely needs that here.

mcolella14 (approved):
LGTM!

ajhenri:
Going to wait on this until the backend rebuild is completed.

PR #305, opened 2022-09-14
[DEVHUB-1435]/[DEVHUB-1436] Next.js ugprade & Use next/image for all images

max-melo:
## Jira Ticket:

[https://jira.mongodb.org/browse/DEVHUB-1435](DEVHUB-1435)
(part of) [https://jira.mongodb.org/browse/DEVHUB-1436](DEVHUB-1436)
## Description:

In pursuit of improved page performance, I wanted to convert all images to use the `next/image` component which has built in image optimization features such as lazy loading, modern image format conversion, etc. There was an issue in Next.js `12.1.0` where using `next/image`'s image component in an app with `trailingSlash: true` caused [endless 308 redirects](https://github.com/vercel/next.js/issues/21337). Because of this, in order to fully adopt `next/image`, I updated Next to `12.3.0`.

One major difference between `12.3.0` and `12.1.0` is the middleware file, which now must live at the root and cannot return a response body ([see here](https://nextjs.org/docs/messages/middleware-upgrade-guide)). Previously we were returning a 500 json response body in the case where a put/post request to an api endpoint had an invalid user agent or origin. This was moved to an explicit api route with a `rewrite` in the middleware.

All images in the site were converted to use `next/image`, except article images which now use [`next/future/image`](https://nextjs.org/docs/api-reference/next/future/image) due to their dynamic nature (`next/image` requires a static width/height on either the image itself or its container). Also, many images that were already using `next/image` were using a custom `thumbnailLoader`, which I removed because specifying a custom loader circumvents Next's image optimization.
 


src/service/get-featured-for-content.test.ts
```diff
@@ -3,6 +3,11 @@ import { Featured, FeaturedItem } from '../interfaces/featured';
 import { getFeaturedForContent } from './get-featured-for-content';
 import * as getFeaturedForTopicModule from './get-featured-for-topic';
 
+jest.mock('./get-featured-for-topic', () => ({
```
max-melo, line 6: Seems like upgrading SWC (a side effect of upgrading next) introduces strange behavior with `jest.spyOn`. [This project](https://github.com/okp4/ui/pull/457) seems to have had a similar issue after upgrading SWC to `1.2.242`. This work-around of mocking the module exports with their actual values was suggested [here](https://github.com/aelbore/esbuild-jest/issues/26#issuecomment-968853688) and it seems to do the trick.

next.config.js
```diff
@@ -21,7 +21,21 @@ const configVals = {
...
+            },
+            {
+                protocol: 'https',
+                hostname: 'mongodb-devhub-cms.s3.us-west-1.amazonaws.com',
+            },
+            {
+                protocol: 'https',
+                hostname: 'i.ytimg.com',
+            },
+        ],
```
ajhenri, line 37: 1) what is `*.mongodb.com` supposed to be for? Is it just for webimages.mongodb.com? If so why not just use that?
2) what is `i.ytimg.com`?

Might be helpful to add comments on that those are supposed to be for.

playwright-tests/pages/index.spec.ts
```diff
@@ -35,7 +35,7 @@ test('Homepage links are correct', async ({ page }) => {
     const allTechLinks = page.locator('a:has-text("View All Technologies")');
     const allProductsLinks = page.locator('a:has-text("View All Products")');
     // There are mobile and desktop links rendered, check both links
-    await allLinksHaveHref(allLangLinks, '/developer/languages/');
-    await allLinksHaveHref(allTechLinks, '/developer/technologies/');
-    await allLinksHaveHref(allProductsLinks, '/developer/products/');
+    await allLinksHaveHref(allLangLinks, '/developer/languages');
+    await allLinksHaveHref(allTechLinks, '/developer/technologies');
+    await allLinksHaveHref(allProductsLinks, '/developer/products');
```
ajhenri, line 40: Just curious why the reason for this change?

ajhenri (commented):
Looks really great! Thanks for doing all this work. Excited to finally be on 12.3.0 as well.

All I have is a couple of questions:

next.config.js
max-melo: `**.mongodb.com` covers `webimages.mongodb.com` and `www.mongodb.com` -- some images are being served from the latter. also more generally i figured that all images coming from any `mongodb.com` domain should be trusted.

`i.ytimg.com` is for youtube thumbnails. next will throw an error on most of the videos pages if this isn't here.

i'll add comments for both

playwright-tests/pages/index.spec.ts
max-melo: after removing the `thumbnailLoader` from all of the `Image` components, a lot of them were giving 404s. i figured out that this has to do with an unnecessary trailing slash that was being added in a util file here:

[https://github.com/mongodb/devcenter/pull/305/files#diff-02af2aabafccca1d0923edaea39e1c8e63813a2e8097f123795ccb850e876ab6L25](https://github.com/mongodb/devcenter/pull/305/files#diff-02af2aabafccca1d0923edaea39e1c8e63813a2e8097f123795ccb850e876ab6L25)

this change fixed the images and allowed them to load, but the tests all had hardcoded trailing slashes when checking the hrefs, so that's why i removed them from all of the related tests

next.config.js
max-melo: this also has me thinking, i wonder if we should add some sort of extra configurability here, maybe via an environment variable. if any different content is added through the cms (e.g. a video from a service other than youtube), we'll have to deploy a change to the codebase to allow the new domain and it'll break in the meantime. 

not sure if this is something that we expect to happen, but just a thought

mcolella14 (commented):
Looks awesome! Love bumping up the Next.js versions, fun to see what they do. Just had one nitpick, and do want to resolve the trailing slash conversation before approving, but everything else looks great.

src/utils/format-url-path.ts
```diff
@@ -22,10 +22,6 @@ export const getURLPath = (url: string | undefined) => {
         }
     }
 
-    if (urlPath[urlPath.length - 1] !== '/') {
-        urlPath += '/';
-    }
```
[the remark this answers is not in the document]
ajhenri: Yeah, this is part of that. We shouldn't be removing this. The reason being-- the topic dropdown will not have the links appended with `/` and instead will be redirects.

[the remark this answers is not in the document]
ajhenri: @max-melo ^

playwright-tests/pages/index.spec.ts
ajhenri: Rather than updating this to remove the `/` here, can we modify the utils below so that:
```
    if (urlPath[urlPath.length - 1] !== '/' && !urlPath.startsWith('_next/images')) {
        urlPath += '/';
    }
```
Or whatever `!urlPath.startsWith('_next/images')` should be so that the images don't end up being 404s?

next.config.js
ajhenri: > this also has me thinking, i wonder if we should add some sort of extra configurability here, maybe via an environment variable. if any different content is added through the cms (e.g. a video from a service other than youtube), we'll have to deploy a change to the codebase to allow the new domain and it'll break in the meantime.

If we expect these to change often, then that sounds like a great idea!

src/utils/format-url-path.ts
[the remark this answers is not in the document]
max-melo: added this back and couldn't reproduce the 404s--not sure what happened but it seems like it was a separate issue 🤷

next.config.js
```diff
@@ -17,11 +17,61 @@ const basePath = '/developer';
...
+        }
+    }, []);
+};
+
+/*
+    Set the NEXT_IMAGE_HOSTS env variable to a semicolon-delimited list of hosts to add extra image hosts
+    e.g. export NEXT_IMAGE_HOSTS="https://google.com/;http://**.test.com/test?abc=123" will generate the entries
+    [{ 'protocol': 'https', 'hostname': 'google.com' }, { 'protocol': 'http', 'hostname': '**.test.com', 'pathname': '/test' }]
+*/
+const IMAGE_PATTERNS = buildImagePatterns(process.env.NEXT_IMAGE_HOSTS);
```
ajhenri, line 53: Cool! Thanks for adding this. I think `NEXT_IMAGE_HOSTS` would need to be set in the drone pipeline, as we set our other env vars at build time. I guess we can use a drone secret for now (though one day I will want to find a way to not use drone for this). Let me know if you need help setting this up. https://github.com/mongodb/devcenter/blob/6a60b966891605493c045d893b3c7b97992df675/.drone.yml#L15-L37

ajhenri (commented):
Sorry a couple more comments, mainly around the `NEXT_IMAGE_HOSTS` environment variable-- but also, I noticed that the search images (loader graphic and "no results" graphic) are not showing up for me locally (they are broken images). Not sure if it's from the 404 issue you pointed our earlier.

ajhenri (approved):
LGTM!

mcolella14 (approved):

PR #348, opened 2022-11-03
[DEVHUB-1546] Consume @mdb/devcenter-components for filter components.

mcolella14:
## Jira Ticket:

[DEVHUB-1546](https://jira.mongodb.org/browse/DEVHUB-1546)

## Description:

Replaces all functionality of the `FilterGroup`, `FilterTag`, and `FIlterTagSection` components with the corresponding components from `@mdb/devcenter-components`. 

Since the shared components had no need for a radio filter input (and since this depends on a feature not yet in the main flora release), I modified the existing `FilterGroup` component to become `RadioFilterGroup`, fulfilling the specific need we have. Once the radio button functionality is available in Flora, this may be added to `@mdb/devcenter-components`.

Everything should functionally and visually be the same in the app.


.eslintrc.json
```diff
@@ -1,3 +1,3 @@
 {
-    "extends": "next/core-web-vitals"
+    "extends": ["next/core-web-vitals", "plugin:react/jsx-runtime"]
```
mcolella14, line 2: Was having some issues with eslint and parsing the `sx` prop with the Theme UI jsx, this seems to help.

src/components/search-filters/desktop/filters.tsx
```diff
@@ -13,64 +14,88 @@ const DesktopFilters: React.FunctionComponent<FiltersProps> = ({
     contentTypeItems = [],
     codeLevelItems = [],
 }) => {
+    const onToggle = (checked: boolean, filter: FilterItem) => {
+        if (checked) {
+            if (filter.subFilters && filter.subFilters.length) {
```
justinmckenzie, line 19: I think you might be able to shorten this to `if (filter?.subFilters.length)`

src/components/search-filters/mobile/filters.tsx
```diff
@@ -33,9 +36,33 @@ const MobileFilters: React.FunctionComponent<MobileFiltersProps> = ({
     closeModal,
 }) => {
     const [tempFilters, setTempFilters] = useState<FilterItem[]>(allFilters);
-    const onTempFilter = (filters: FilterItem[]) => {
-        setTempFilters(filters);
+    const onToggle = (checked: boolean, filter: FilterItem) => {
+        if (checked) {
+            if (filter.subFilters && filter.subFilters.length) {
```
justinmckenzie, line 41: same here

src/hooks/search/index.ts
```diff
@@ -130,13 +130,13 @@ const useSearch = (
...
                         l1.name === item ||
-                        l1.subItems.find(l2 => l2.name === item)
+                        l1.subFilters?.find(l2 => l2.name === item)
                 );
                 if (filterProduct) {
                     if (filterProduct.name !== item) {
                         // This means it's an L2 match.
                         filtersList.push(
-                            filterProduct.subItems.find(
+                            filterProduct.subFilters?.find(
```
justinmckenzie, line 139: Guess it technically doesn't matter, but wouldn't you want to chain it on the first one because you might still run into the chance of `subFilters` being undefined, `filterProduct?.subFilters.find(...)`

src/hooks/search/utils.ts
```diff
@@ -173,8 +184,8 @@ export const getFilters = async (
                     const filterItem: FilterItem = {
                         type: tag.type,
                         name: tag.name,
-                        count: allFilters || type === contentType ? 1 : 0,
-                        subItems: [],
+                        // count: allFilters || type === contentType ? 1 : 0,
```
justinmckenzie, line 187: comment

src/hooks/search/utils.ts
```diff
@@ -185,37 +196,42 @@ export const getFilters = async (
...
+            item.subFilters.sort(
+                (prev, next) => (next.count || 0) - (prev.count || 0)
+            );
         }
     });
 
     const l1Items = filterItems
         .filter(({ type }) => type === 'L1Product')
-        .sort((prev, next) => next.count - prev.count);
+        .sort(sortFunction);
```
justinmckenzie, line 212: 👍

justinmckenzie (approved):
optional code cleanup, lgtm

ajhenri (approved):
Looks great. Very cool that we have our own component package now.

src/hooks/search/index.ts
mcolella14: So if I'm [reading this right](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining), I think it should still be `subFilters?.` because the optional chain checks to see if the property with the `?` appended is undefined before trying to access a further property of it. `filterProduct` will always be defined, but `filterProduct.subFilters` might not be so we add the `?` to `subFilters`?

justinmckenzie: oh word okay my mistake then, I always figured you would want to first check the property existing and if it exists, then you can safely call the method on it. Thanks for clarifying and ignore any other comments of mine lol

mcolella14: lol all good, i feel like they could have done a better job making optional chaining more intuitive, I'm always looking at that page for reference.

PR #360, opened 2022-11-14
[DEVHUB-1575] Search & Content Type Page Refactor

max-melo:
## Jira Ticket:

[[DEVHUB-1575] Content Type Page Refactor for Events Page](https://jira.mongodb.org/browse/DEVHUB-1575)

## Description:

This started as a small project to add some flexibility to the Content Type page to assist the new Events page (whose layout is different than the rest of the content type pages) but gradually scope creeped into a pretty big refactoring effort. Hopefully, the resulting components here should be easier to read and be more performant (especially search, see the screen recordings).

I took a lot of liberties with the decisions I made here, so I'm going to write several comments below explaining some of the them.  Feel free to disagree with any of the choices while reviewing!

Some of the main changes:
- Split search into a few sub-components which are wired together with useSearch for added flexibility (search box can be positioned away from sort box/results while still sharing state via `useSearch`)
- Refactored `useSearch` to export specific props objects meant to be wired into each search component (e.g. exports `searchBoxProps` to be spread into the SearchBox component)
- Moved logic about page metadata into a hook `useSearchMeta`
- Optimized search and filtering to remove unnecessary re-renders which noticeably speeds up the searching user experience
- Converted all pages utilizing search to use the same components/hook and removed duplicate code
- Refactored content type page to be much less verbose and to allow a custom "body" which will facilitate the events page
- Generalized filter items so that the new events filters can be added more easily

Notice how the current search is "jerky" in that you type some characters in, the page freezes, and then they all show up at once, whereas the new search is smooth and fast.

Current search (prod)

https://user-images.githubusercontent.com/110849018/201768530-d8a112e6-7ade-4f10-89a2-d6ebb63c5526.mov

New search

https://user-images.githubusercontent.com/110849018/201768542-6a6dd63b-2255-4f05-89eb-34d2ed149030.mov


src/components/search-filters/desktop/filters.tsx
```diff
@@ -1,105 +1,58 @@
...
-    contentTypeItems = [],
-    codeLevelItems = [],
-}) => {
-    const onToggle = (checked: boolean, filter: FilterItem) => {
-        if (checked) {
-            if (filter.subFilters?.length) {
-                const subFiltersToAdd = filter.subFilters.filter(
-                    subFilter =>
-                        !allFilters.find(
+const DesktopFilters: React.FunctionComponent<FiltersProps> = memo(
```
max-melo, line 8: I memo'd a few of these components related to this work because it actually does seem to cut down on unnecessary re-renders according to the React profiler 
Screen Shot 2022-11-14 at 7 30 30 PM

src/components/search-filters/types.ts
```diff
@@ -14,12 +14,6 @@ export interface RadioFilterGroupProps {
...
-    allFilters: FilterItem[];
-    l1Items: FilterItem[];
-    languageItems: FilterItem[];
-    technologyItems: FilterItem[];
-    contributedByItems: FilterItem[];
-    expertiseLevelItems: FilterItem[];
-    contentTypeItems?: FilterItem[];
-    codeLevelItems?: FilterItem[];
+    filters: FilterItem[];
+    filterItems: { key: string; value: FilterItem[] }[];
```
max-melo, line 18: Converted all of these specifically named filters to this `filterItems` prop so filters can be added more easily. Array of objects instead of just an object because the order matters.

src/hooks/search/index.ts
```diff
@@ -47,67 +51,93 @@ const useSearch = (
 
     // TODO: Refactor to useSWRInfinite and implement client-side pagination.
     const { data, error, isValidating } = useSWR(key, fetcher, {
-        revalidateIfStale: false,
+        // revalidateIfStale: false,
         revalidateOnFocus: false,
         revalidateOnReconnect: false,
         shouldRetryOnError: false,
+        fallback: {
```
max-melo, line 58: There previously was some verbose code related to setting initial data for SEO purposes, but I found this `fallback` option on the [Vercel docs](https://swr.vercel.app/docs/with-nextjs#pre-rendering-with-default-data) and it seems to have the same performance/SEO benefits while allowing for much cleaner code. This data seems to be statically generated in the build output so it shouldn't break SEO, but let me know if I'm missing anything.

src/hooks/search/index.ts
```diff
@@ -218,39 +248,47 @@ const useSearch = (
...
-        setAllFilters,
-        onSearch: debouncedOnSearch,
-        onFilter,
-        searchString,
-        setSearchString,
-        fullyLoaded,
-        numberOfResults,
-        onSort,
-        sortBy,
+        searchBoxProps: {
```
max-melo, line 272: Put a lot of thought into the best way to refactor the search so it could be used on the topic pages as well as content type/sitewide search. The crux of the problem is to find a way where the search logic could be centralized but applied to multiple components that might be scattered throughout the page. This seems to be the best way, which was similar to the old implementation, but I thought packaging the props by component is probably the cleanest so that they can later be spread into each one easily. Also, the case in which a component handles multiple search functionalities (e.g. mobile filter component handling sorting & filtering) becomes much more declarative -- you just need to spread in the related prop group (`{...sortBoxProps} {...filterProps}`) 

Also, by not relying on one centralized search component, and rather using `useSearch` in combination with whichever specific search related components are needed, it removes a lot of the need for excessive prop passing and one-off if-statements.

src/hooks/search/meta.ts
```diff
@@ -0,0 +1,76 @@
+import getConfig from 'next/config';
+import { useRouter } from 'next/router';
+import { useCallback, useState } from 'react';
+import { getCanonicalUrlWithParams, getMetaDescr } from '../../utils/seo';
+import { replaceHistoryState } from './utils';
+
+export const useSearchMeta = (
```
max-melo, line 7: this hook is pretty self explanatory but I was seeing a lot of duplicated page-meta-related code and it tidies things up greatly to have that stuff centralized here

src/components/search/results.tsx
```diff
@@ -1,49 +1,72 @@
...
+        pageNumber,
+        updatePageMeta = () => {},
+        noResultsFooter = DefaultNoResultsFooter,
+        layout = 'list',
+        extraStyles = {},
+    }) => {
+        const maxPage = Math.ceil(
+            Math.max(results.length, 1) / DEFAULT_PAGE_SIZE
+        );
+        const [currentPage, setCurrentPage] = useState(
```
max-melo, line 34: Previously the pagination related stuff was in `useSearch` but, in my opinion, it makes a bit more sense for `useSearch` to only be concerned with everything leading up to the search & fetching the data. Pagination, on the other hand, isn't a parameter for executing the search and I think only the component responsible for displaying the results should be concerned with it.

src/page-templates/content-type/content-type.tsx
```diff
@@ -1,421 +1,138 @@
...
+const heroCrumbs = [
+    {
+        text: 'MongoDB Developer Center',
+        url: '/',
+    },
+];
+
+// Debug for DEVHUB-1501 which is not yet replicable.
+// If data is empty, capture an exception for Sentry.
+const useEmptyDataDebug = (results: any, error: any) => {
```
max-melo, line 44: There was more to this debug code previously but since we're supplying a fallback to `useSWR` now it should be guaranteed that the data is non-empty.... but if for some reason it still is we still capture a Sentry report.

src/page-templates/content-type/content-type.tsx
```diff
@@ -424,166 +141,183 @@ const ContentTypePage: NextPage<ContentTypePageProps> = ({
...
+                                {...filterProps}
+                                {...sortBoxProps} // Mobile filters include sorting
+                                filterItems={filterItems}
+                                closeModal={() => setMobileFiltersOpen(false)}
+                            />
+                        )}
+                    </div>
+
+                    {!children && (
+                        <div sx={searchWrapperStyles}>
```
max-melo, line 179: Using a flexbox here to deduplicate some markup (had to duplicate the sortbox before when we were using a grid layout based on if the featured section was showing)

src/page-templates/content-type/content-type.tsx
```diff
@@ -424,166 +141,183 @@ const ContentTypePage: NextPage<ContentTypePageProps> = ({
...
-                        )}
-                    </div>
+                                        Back to all {contentType.toLowerCase()}s
+                                    </Button>
+                                }
+                            />
+                        </div>
+                    )}
+
+                    {children && children(searchProps, searchMetaProps)}
```
max-melo, line 312: This is similar to a render prop to render a completely different "body" to the page while still borrowing some of the wrappers and the searching logic. This'll be needed for the events page which has a custom multi-tiered search.

mcolella14 (approved):
Awesome work with this, I tested this locally and the the search input and filters feel so much better without the rerenders. Just really had the one comment about pagination re: the new backend, but nothing blocking. Thanks for taking this on, makes a huge difference!

src/components/search/search-box.tsx
```diff
@@ -0,0 +1,67 @@
+import { useState } from 'react';
```
justinmckenzie, line 1: double import here and line 5 from `react`

justinmckenzie, line 1: double import here and line 5 from `react`

src/components/search-filters/mobile/filters.tsx
```diff
@@ -1,4 +1,5 @@
-import { useState } from 'react';
+import { memo } from 'react';
+import { useCallback, useState } from 'react';
```
justinmckenzie, line 2: double import

src/components/search-filters/desktop/filters.tsx
```diff
@@ -1,105 +1,58 @@
+import { memo } from 'react';
 import { FiltersProps } from '../types';
 
 import { FilterGroup, FilterItem } from '@mdb/devcenter-components';
+import { CONTENT_TYPE_NAME_MAP } from '../../../data/constants';
+import { useCallback } from 'react';
```
justinmckenzie, line 6: double import

src/hooks/search/index.ts
```diff
@@ -47,67 +51,93 @@ const useSearch = (
 
     // TODO: Refactor to useSWRInfinite and implement client-side pagination.
     const { data, error, isValidating } = useSWR(key, fetcher, {
-        revalidateIfStale: false,
+        // revalidateIfStale: false,
```
justinmckenzie, line 54: commenting, just in case this was an oversight

justinmckenzie (approved):
just code cleanup stuff, nice job

src/components/search/results.tsx
max-melo: ah okay I wasn't aware of this, good catch! I'll add a comment

src/hooks/search/index.ts
max-melo: so this was intentionally commented, but what's going on here is a bit fuzzy. my understanding is this:
- the `fallback` i.e. initial search data is only the first "page" of data and is baked into the build files by next
- when the page is loaded, `useSWR` initially returns the `fallback` data, but simultaneously runs the query to fetch the full up-to-date data
- `useSWR` switches the `fallback` data to the new up-to-date data once the query completes, which usually is very soon after the page loads

all of this is to say that if `revalidateIfStale` is false, it will never perform that initial search query and only the first page of data will be shown until the user interacts with one of the search components, effectively hiding the "Load more" button on first render. commenting this out doesn't seem to have any other side effects from what i can tell.

src/components/search/results.tsx
ajhenri: Nice catch @mcolella14, thanks for explaining here. @max-melo While the new backend can use pagination, it isn't required. If you do not pass a page number, the search will return all results still -- so as Matt mentioned, only a comment is needed.

ajhenri:
@max-melo This is great, and one very difficult refactor-- thanks for doing this. I need a bit more time to look as this is a hefty PR but should have completed by EOD today.

src/hooks/search/index.ts
```diff
@@ -218,39 +248,47 @@ const useSearch = (
...
+            sortBy,
+        },
+        resultsProps: {
+            results: filteredData,
+            error,
+            isValidating,
+            searchString,
+            filters,
+        },
+        clearAll,
```
ajhenri, line 291: You need to return `pageNumber` here also, in case there is a page number passed within getServerSideProps.
It is causing the issue where we visit the page http://localhost:3000/developer/languages/javascript/?page=6 and then click "Load More" and it becomes http://localhost:3000/developer/languages/javascript?page=2.

src/page-templates/topic-content-type-page/topic-content-type-page-template.tsx
```diff
@@ -246,25 +283,53 @@ export const TopicContentTypePageTemplate: NextPage<
...
+                        {contentType === 'Code Example' && (
+                            <ExtraCodeExampleCheckboxes {...filterProps} />
                         )}
-                        sx={spanAllColumns}
-                    />
+
+                        <SearchResults
+                            {...resultsProps}
+                            pageNumber={pageNumber}
+                            slug={topicSlug}
```
ajhenri, line 324: Loading more on the topic page is resulting in a missing slash which could impact SEO
http://localhost:3000/developer/languages/javascript?page=2
Should be
http://localhost:3000/developer/languages/javascript/?page=2

src/components/search/results.tsx
```diff
@@ -54,15 +80,105 @@ const Results: React.FunctionComponent<ResultsProps> = React.memo(
...
+                                                        'inc70',
+                                                        null,
+                                                        'inc90',
+                                                    ],
+                                                }}
+                                            >
+                                                {!isValidating && results && (
+                                                    <a
+                                                        href={loadMoreHref}
+                                                        onClick={onLoadMore}
```
ajhenri, line 136: I don't think this is behaving the way we want it to on every page.

An example of what's broken: When I go to http://localhost:3000/developer/languages/javascript/code-examples/ and click load more, it is going to an entirely different page http://localhost:3000/developer/languages/javascript?page=2.

I can't tell yet what the cause is, but wondering if it has to do with the memoization?

src/components/search/index.ts
```diff
@@ -1,3 +1,5 @@
-import Search from './search';
+import SearchResults from './results';
+import SearchBox from './search-box';
+import SortBox from './sort-box';
 
-export default Search;
+export { SearchResults, SearchBox, SortBox };
```
ajhenri, line 5: We definitely needed to do this, as the search becomes more complex. Great work!

max-melo:
@ajhenri Thanks for catching these issues. Should all be fixed now, but definitely worth a second pair of eyes re-checking as there's a lot of complex logic around this query param stuff that I didn't realize before, so I want to be 100% sure I didn't miss anything. 
Somewhat related, I also refactored some of the sorting code due to some subtle issues I noticed--previously navigating to a `?page=#` and then sorting would only sort what's on that page and not reset the page--but hopefully that should be fixed now.

ajhenri:
>@ajhenri Thanks for catching these issues. Should all be fixed now, but definitely worth a second pair of eyes re-checking as there's a lot of complex logic around this query param stuff that I didn't realize before, so I want to be 100% sure I didn't miss anything.
Somewhat related, I also refactored some of the sorting code due to some subtle issues I noticed--previously navigating to a ?page=# and then sorting would only sort what's on that page and not reset the page--but hopefully that should be fixed now.

Just spoke with @max-melo out of band. All the fixes look good! The only item pending is the other SEO related thing (regarding JS-disabled pages).

ajhenri (approved):

PR #366, opened 2022-11-17
[CHORE for DEVHUB-1531]: Refactor & CleanUp of Page Templates Ahead of Events Content Type

justinmckenzie:
## Jira Ticket:

Link to jira ticket [[DEVHUB-1531](https://jira.mongodb.org/browse/DEVHUB-1531)]

## Description:

- Cleaned up the `page-templates` directory and changed some folder/file names to keep things consistent and detail relationships to the other templates (if anyone disagrees with these changes, happy to revert them, just thought it made things a little cleaner)
- Refactored `content-page-template` to render `<VideoOrPodcastContent />` or `<DefaultContent />` based on the Content Type. This helps minimize the usage of the `vidOrPod` conditional to only once. Tried to remove as much as I could to make the file a little less daunting, there's probably some more stuff that can be separated. 
- Extracted `requestContentModal` local state in all of the templates to `useRequestModalContext` that is now consumed by the separate templates


justinmckenzie:
changing some directory names probably makes this PR a bit of a pain to review...majority of the changes are in `page-templates/content-page/content-page-template.tsx` (new) and `page-templates/main-content-page/content-page-template.tsx` (old). I can revert the directory changes for the time being so the files can be compared side by side, let me know

src/pages/_app.tsx
```diff
@@ -55,11 +56,13 @@ function MyApp({ Component, pageProps, session }: AppProps & CustomProps) {
...
-                                    <Component {...pageProps} />
-                                </ErrorBoundary>
-                            </Layout>
+                            <RequestContentModalProvider>
+                                <Layout pagePath={pagePath}>
+                                    <ErrorBoundary>
+                                        <Component {...pageProps} />
+                                    </ErrorBoundary>
+                                </Layout>
+                            </RequestContentModalProvider>
```
mcolella14, line 65: Is there any overhead to wrapping the whole app with this? Just because there are a lot of routes that don't have the request content functionality, and I'm wondering if it might make more sense to just wrap the page templates that do have that functionality instead?

src/page-templates/content-page/content-page-template.tsx
```diff
@@ -0,0 +1,503 @@
...
+                    sx={styles.requestBtn}
+                    onClick={() => setModalStage('text')}
+                >
+                    {requestButtonText}
+                </Button>
+            )}
+        </div>
+    );
+
+    function VideoOrPodcastContent() {
```
mcolella14, line 254: Very nitpicky, but any reason for using these function declarations instead of arrow function? Just pointing it out since there are already a lot of arrow functions in this file.

mcolella14 (approved):
Looks great to me! Just some nitpicks, but I think this'll make things a lot easier to work with.

src/pages/_app.tsx
justinmckenzie: thats a great point, ill look into moving it elsewhere

justinmckenzie: so i looked into this a little bit more...the issue is that `<ContentType />` is returned separately 7 times, plus adding the other usages, I would probably have to wrap `<RequestContentModalProvider />` 10+ times, which I would say mitigates the purpose of trying to use this context. If we think using it in context doesn't provide any value to the app, I can revert it back to local state no problem

PR #370, opened 2022-11-17
Fix Canonical URL Logic

max-melo:
## Jira Ticket:

N/A 

## Description:

Found a small bug while poking around staging, when a search page reset its page to 1, the canonical url would reset to `/developer/<slug>/?page=1` instead of just `/developer/<slug>`. This should fix that


ajhenri (approved):
LGTM

PR #381, opened 2022-11-30
[DEVHUB-1559]: Integrate Google Places API 

justinmckenzie:
## Jira Ticket:

Link to jira ticket [[DEVHUB-1559](https://jira.mongodb.org/browse/DEVHUB-1559)]

## Description:

Adds Location Search Component for upcoming Events implementation

### DIFF
--- a/src/components/location-search/location-search.tsx
+++ b/src/components/location-search/location-search.tsx
@@ -1,0 +1,118 @@
+import debounce from 'lodash.debounce';
+import { ComboBox, ESystemIconNames } from '@mdb/flora';
+import { ChangeEvent, useCallback, useMemo, useState } from 'react';
+
+import { getURLPath } from '../../utils/format-url-path';
+
+type Option = string | { icon: string; label: string };
+
+// TODO: needs to be updated with icon names that will be added by Web Team - https://jira.mongodb.org/browse/WEBSITE-13740
+const DEFAULT_OPTIONS = [
+    { icon: ESystemIconNames.ARROW_UP, label: 'Current Location' },
+    { icon: ESystemIconNames.ARROW_DOWN, label: 'Virtual' },
+];
+
+interface LocationSearchProps {
+    onSelectCallback?: (selection: string) => void;
+}
+
+export default function LocationSearch({
+    onSelectCallback = () => null,
+}: LocationSearchProps) {
+    const [value, setValue] = useState('');
+    const [isLoading, setIsLoading] = useState(false);
+    const [options, setOptions] = useState<Option[]>(DEFAULT_OPTIONS);
+
+    async function geolocationSuccessCb({
+        coords: { latitude, longitude },
+    }: {
+        coords: GeolocationCoordinates;
+    }) {
+        const req = await fetch(
+            `${getURLPath('api/geocode')}?latlng=${latitude},${longitude}`
+        );
+        const { error, results } = await req.json();
+
+        if (error) {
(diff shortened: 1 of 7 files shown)


justinmckenzie:
this should only stay in staging until events is shipped out

src/components/location-search/location-search.tsx
```diff
@@ -0,0 +1,118 @@
...
+const DEFAULT_OPTIONS = [
+    { icon: ESystemIconNames.ARROW_UP, label: 'Current Location' },
+    { icon: ESystemIconNames.ARROW_DOWN, label: 'Virtual' },
+];
+
+interface LocationSearchProps {
+    onSelectCallback?: (selection: string) => void;
+}
+
+export default function LocationSearch({
```
max-melo, line 19: i think we may want to consider moving some of this logic, especially state/value stuff, to `useSearch` for the sake of consistency, since the other three search components' searching logic resides there currently. 

also, if i'm reading the code right, i don't believe there will be a way to clear the location search from the parent component, which would be necessary (i think? can't find a figma but that's the way it is on other content type pages) for the no results found "Back to Events" button on the events content type page. e.g. a line could be added in this function https://github.com/mongodb/devcenter/blob/68e1df61db432b891e0b4f97b21a9f043492ab63/src/hooks/search/index.ts#L252 to facilitate clearing

src/pages/api/geocode.ts
```diff
@@ -0,0 +1,22 @@
+import type { NextApiRequest, NextApiResponse } from 'next';
+import { withSentry } from '@sentry/nextjs';
+
+async function geocodeHandler(req: NextApiRequest, res: NextApiResponse) {
+    try {
+        const request = await fetch(
+            `https://maps.googleapis.com/maps/api/geocode/json?latlng=${req.query?.latlng}&result_type=locality&key=${process.env.GOOGLE_PLACES_API_KEY}`
```
max-melo, line 7: do we want to add a short circuit here to check the value of `req.query`? using optional chaining here might result in unnecessary fetches to `.../geocode/json?latlng=undefined&...`. might be a non-issue if `req.query` is never null

src/components/location-search/location-search.tsx
```diff
@@ -0,0 +1,118 @@
...
+                setOptions(
+                    results.map(
+                        ({ description }: { description: string }) =>
+                            description
+                    )
+                );
+            }
+
+            setIsLoading(false);
+        }, 500),
```
max-melo, line 74: this debounce ms value differs from the one we use in the search field [here](https://github.com/mongodb/devcenter/blob/68e1df61db432b891e0b4f97b21a9f043492ab63/src/hooks/search/index.ts#L99). might be worth moving this to a constant and sharing it between the two.

src/components/location-search/location-search.tsx
justinmckenzie: yeah thats a fair point, i built out this component in isolation for now just to know the Google API's and the ComboBox were working. ill look into how this will tie in to the existing search functionality

mcolella14 (approved):
LGTM!

gfunk1230 (approved):

max-melo (approved):

justinmckenzie:
Merging for now so I can close out related tickets to this work [DEVHUB-1467](https://jira.mongodb.org/browse/DEVHUB-1467) and [DEVHUB-1572](https://jira.mongodb.org/browse/DEVHUB-1572), will incorporate @max-melo's `useSearch` comments in a new PR

PR #393, opened 2022-12-09
[DEVHUB-1589]: Code Clean Up Pt. 1 - Enhance Linter & Clean Up Errors/Warnings

justinmckenzie:
## Jira Ticket:

Link to jira ticket [[DEVHUB-1589](https://jira.mongodb.org/browse/DEVHUB-1589)]

## Description:

- Expanded `.eslintrc.json` to be stricter and cleared up all errors and warnings that came as a result. Only remaining warnings are ` Warning: Unexpected any. Specify a different type.  @typescript-eslint/no-explicit-any` which will be a pretty heavy lift to clear all of those.
- Removed all unused exports
- Some code cleanup and small refactors along the way


.eslintrc.json
```diff
@@ -1,3 +1,27 @@
...
+        "plugin:@typescript-eslint/eslint-recommended",
+        "plugin:@typescript-eslint/recommended",
+        "prettier"
+    ],
+    "rules": {
+        "@typescript-eslint/no-unused-vars": [
+            "error",
+            { "ignoreRestSiblings": true }
+        ],
+        "react-hooks/exhaustive-deps": "off"
```
max-melo, line 25: are we sure about turning this one off? feels like it's bad practice to ignore these warnings unless there's a really good reason not to. in my experience, it can lead to some hard-to-diagnose bugs with stale values and it usually suggests that there's a shortcoming in the way the component code was designed. and, if the dev really needs to, they can just write an eslint disable comment, which has the added benefit of forcing them to consider their decision more concretely and sometimes prompting another comment to explain why. just my 2 cents

max-melo: not sure if it was even on previously or not but if we're making the eslintrc stricter i think it's a good idea to turn it on personally

justinmckenzie: i can turn it back on, TBH the warning was only occurring within a couple hooks that I was unfamiliar with so I was hesitant about adding to the dependency array and it producing unwanted side effects. ill remove this and let the devs who are familiar with the implementation add to it as they come across it in the future

max-melo (approved):

mcolella14 (approved):

PR #396, opened 2022-12-15
[DEVHUB-1545] Updated featured card sizing

amybguo:
## Jira Ticket:

Link to jira ticket [DEVHUB-1545](https://jira.mongodb.org/browse/DEVHUB-1545)

## Description:

Changed card sizing in featured card section

-   [ ] I have performed a self-review(QA) of my code and reviewed with Design or Product (If Applicable)?

before:
Screen Shot 2022-12-15 at 3 53 11 PM

after:
Screen Shot 2022-12-15 at 3 53 32 PM

### DIFF
--- a/src/components/card-section/types.ts
+++ b/src/components/card-section/types.ts
@@ -11,4 +11,5 @@ export interface FeaturedCardSectionProps {
     content: ContentItem[];
     title?: string;
     className?: string;
+    featuredCardType?: string;
 }
--- a/src/components/card-section/types.ts
+++ b/src/components/card-section/types.ts
@@ -11,5 +11,5 @@ export interface FeaturedCardSectionProps {
     content: ContentItem[];
     title?: string;
     className?: string;
-    featuredCardType?: string;
+    featuredCardType: 'middle' | 'large';
 }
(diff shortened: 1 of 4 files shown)


mcolella14 (changes requested):
Ok so this might be a bigger change. Basically we use this `FeaturedCardSection` component in 2 places: the Content Type page (like https://www.mongodb.com/developer/articles/) and the Topic page (like https://www.mongodb.com/developer/languages/javascript/). We want to have this change take effect only on the Content Type page, but not the Topic page. So essentially we will have to add a prop to this component to choose whether to use the big cards or the little cards.

src/components/card-section/types.ts
```diff
@@ -11,4 +11,5 @@ export interface FeaturedCardSectionProps {
     content: ContentItem[];
     title?: string;
     className?: string;
+    featuredCardType?: string;
```
mcolella14, line 14: I think it might be best to explicitly specify the options here. So like
```typescript
featuredCardType: 'middle' | 'large';
```
and then just default that prop to `large`
```typescript
({ content, className, title = 'Featured', featuredCardType = 'large' })
```

It just helps to make more obvious how we can use that prop.

mcolella14 (commented):
Looks great! Just had the one comment about the prop being more explicit.

mcolella14 (approved):
Looks great!

PR #401, opened 2022-12-20
[DEVHUB-1332, DEVHUB-1548] Add auto-sized CodeBlock and Terraform logo.

mcolella14:
## Jira Ticket:

[DEVHUB-1548](https://jira.mongodb.org/browse/DEVHUB-1548)
[DEVHUB-1332](https://jira.mongodb.org/browse/DEVHUB-1332) (this was already closed as it was a Flora contribution ticket).

## Description:

Upgraded Flora to 1.3.1 to include the Terraform logo. Added in the Code Block auto height functionality. Added the Terraform logo.


src/components/article-body/body-components/codeblock.tsx
```diff
@@ -10,15 +12,32 @@ export const CodeBlock = ({ lang, value }: { lang: string; value: string }) => {
...
+    // This seems to be a CodeMirror issue, but is especially apparent here without line wrapping.
+    // Basically this will force the component to rerender after 1 second, which allows it to recalculate the line length.
+    // https://github.com/codemirror/codemirror5/issues/5639 describes a similar issue.
+
+    const [, setReady] = useState(false);
+    useEffect(() => {
+        setTimeout(() => {
+            setReady(true);
+        }, 1000);
+    }, []);
```
mcolella14, line 26: I'm open to suggestions if someone knows a better way to do this.

justinmckenzie: https://reactjs.org/docs/hooks-faq.html#is-there-something-like-forceupdate

if there's no way to fix this at the Code Mirror level, this forcing of an update might be a _little_ cleaner so we can avoid using `setTimeout` and you can just call the `forceUpdate()` in the hook

src/components/article-body/body-components/codeblock.tsx
```diff
@@ -10,15 +12,32 @@ export const CodeBlock = ({ lang, value }: { lang: string; value: string }) => {
...
     return (
-        <div
-            // Need this because Safari/mobile browsers don't normally render the border radius on scroll containers without this.
-            // https://stackoverflow.com/a/58283449
-            sx={{
-                '.CodeMirror-simplescroll': { isolation: 'isolate' },
-            }}
-        >
-            <CodeSnippet height={height} language={language} code={value} />
+        <div sx={styling}>
```
justinmckenzie, line 34: i touched this in the refactor PR so just a versioning mismatch here, but would you mind switching it back? just so we can avoid creating unnecessary variables

src/components/article-body/body-components/codeblock.tsx
max-melo: seems like this approach doesn't work when the user has a slow internet connection. tested it with "Fast 3G" network throttling and the code is cut off at the end again for me.

not sure exactly what the issue is with codemirror, but is it possible to use either a [ResizeObserver](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) or a [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) to trigger a re-render once whatever change which is causing the problem (either visual or dom tree) is complete?

src/components/article-body/body-components/codeblock.tsx
mcolella14: Ah sorry about that, got confused with the merge conflict, I'll set it back to how you had it, a lot simpler that way.

src/components/article-body/body-components/codeblock.tsx
mcolella14: @max-melo Ah good catch on that, yeah I'm not exactly sure what the Codemirror issue is, but I will look  into those APIs to see if I can determine when the code block has fully rendered and then try to get it to recalculate. 

@justinmckenzie yeah I agree I think that would be a little cleaner, I think I'll try to combine this with Max's suggestion.

src/components/article-body/body-components/codeblock.tsx
```diff
@@ -7,18 +8,68 @@ import {
...
+        }
+
+        return () => {
+            // Clean up and disconnect, needed if the user hasn't scrolled the code block into view.
+            if (observer) {
+                observer.disconnect();
+            }
+        };
+    }, [wrapper]);
+
```
mcolella14, line 53: @justinmckenzie @max-melo So I used IntersectionObserver along with the `forceUpdate` logic to just rerender once when the element is scrolled into view. Seems like the simplest  workaround to this issue at the moment, but let me know what you think.

harikakotipalli: is this when the code block is large ?

harikakotipalli (approved):

max-melo (approved):
LGTM! (besides a couple of spelling nitpicks)

src/components/article-body/body-components/codeblock.tsx
```diff
@@ -7,18 +8,68 @@ import {
 } from '../../../utils/get-code-height';
 
 export const CodeBlock = ({ lang, value }: { lang: string; value: string }) => {
+    // This is a workaround to a bug where a long line of code will be cut off at the end.
+    // This seems to be a CodeMirror issue, but is especially apparent here without line wrapping.
+    // https://github.com/codemirror/codemirror5/issues/5639 describes a similar issue.
+
+    // We're utilizing the Intersection Observer API (https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API)
+    // to force the component to rerender when the Code Block elemtn comes into the viewport.
```
max-melo, line 16: `elemtn` -> `element`

src/components/article-body/body-components/codeblock.tsx
```diff
@@ -7,18 +8,68 @@ import {
...
+
+    const [, forceUpdate] = useReducer(x => x + 1, 0); // https://reactjs.org/docs/hooks-faq.html#is-there-something-like-forceupdate
+    const wrapper = useRef<HTMLDivElement>(null);
+
+    const onVisible = useCallback(
+        (
+            entries: IntersectionObserverEntry[],
+            observer: IntersectionObserver
+        ) => {
+            // When this gets called with the element intersecting the vieport, we rerender and disconnect the observer.
```
max-melo, line 26: `vieport` -> `viewport`

src/components/article-body/body-components/codeblock.tsx
mcolella14: No this can happen on both the `large` and `auto` setting.

justinmckenzie (approved):

gfunk1230 (approved):

PR #406, opened 2023-01-05
[DEVHUB-1531]: Industry Event Details Page

justinmckenzie:
## Jira Ticket:

Link to jira ticket [DEVHUB-1531](https://jira.mongodb.org/browse/DEVHUB-1531)

## Description:

Describe your changes in detail so for example expected behavior, actual behavior, steps tp reproduce the behavior


src/api-requests/get-events.ts
```diff
@@ -0,0 +1,81 @@
...
+            title: name
+            contentDate: originalPublishDate
+        }
+        newVideos: new_videos {
+            title
+            contentDate: originalPublishDate
+        }
+        industryEvents: industry_events {
+            title
+            test:
```
harikakotipalli, line 58: here

src/api-requests/get-industry-events.ts
```diff
@@ -43,6 +43,32 @@ export const industryEventsFields = `
   location
   slug
   start_time
+  registration_url
+  virtual_meetup_url
+  virtual_meetup_url_text
+  related_content {
+    newArticles: new_articles {
+        title: name
+        contentDate: originalPublishDate
```
harikakotipalli, line 52: contentDate for articles should be originalPublishDate | published_at, original publish date wont be there for the ones published from CMS. We had few articles in other world, which were migrated to our CMS, only those have original publish date.

src/page-templates/content-page/content-page-template.tsx
```diff
@@ -364,81 +375,186 @@ const ContentPageTemplate: NextPage<ContentPageProps> = ({
...
+                    {isIndustryEvent && (
+                        <>
+                            <TypographyScale
+                                variant="heading5"
+                                sx={{
+                                    marginTop: 'inc60',
+                                    marginBottom: 'inc40',
+                                }}
+                            >
+                                Speakers
```
harikakotipalli, line 491: do we need Header speakers when there are no speakers ?
image

harikakotipalli (approved):

harikakotipalli:
LGTM!

src/page-templates/content-page/content-page-template.tsx
justinmckenzie: good catch, ill adjust

src/api-requests/get-industry-events.ts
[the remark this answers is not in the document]
justinmckenzie: updated

src/page-templates/content-page/content-page-template.tsx
```diff
@@ -364,81 +375,186 @@ const ContentPageTemplate: NextPage<ContentPageProps> = ({
             2,
             -1
         );
+        let displayHeaderImage = true;
+        // Industry Events do not have to show a placeholder image
+        if (isIndustryEvent && !image?.url) {
```
max-melo, line 380: Can this just be `const displayHeaderImage = !industryEvent || image?.url`?

max-melo: also maybe I'm missing something, but is the logic here that images on industry events are optional, or that they are never shown no matter what? if it's the former, then it seems that we're missing some lines in the gql to grab the image field. the reason i want to confirm this logic is because i tested on my local and navigated to an industry event that had an image in strapi, but no image showed up on the industry event page.

max-melo:
small styling nitpick, but on certain screen sizes the social buttons look like they're touching the tags. probably just a bit of  padding is needed here
Screen Shot 2023-01-13 at 11 05 35 AM

max-melo:
another styling nitpick, if the virutal link text is a long unbroken string, it seems to break the layout
Screen Shot 2023-01-13 at 11 13 23 AM

this may not matter, but I could envision maybe a URL being put into strapi that could break the layout in this way. a `word-wrap: break-word` should fix this

src/page-templates/content-page/content-page-template.tsx
justinmckenzie: ah crap good catch, I missed adding it when I was merging our two queries together

src/api-requests/get-industry-events.ts
```diff
@@ -43,6 +43,33 @@ export const industryEventsFields = `
   location
   slug
   start_time
+  registration_url
```
max-melo, line 46: github won't let me add a comment earlier in the file, but I think this query is missing the `title` field under `authors`. i noticed that the authors' titles were missing on my local. looks like the code is there in the JSX to display it, it just needs to be grabbed from here.

max-melo (approved):

PR #412, opened 2023-01-13
Fixes for Timestamps and FIlters on Events page

max-melo:
## Description:
Adding in a couple of quick fixes for some of the issues noted on the events page. Once we have time we should probably come up with a more robust way of transforming API responses so we don't have to make these same checks in multiple parts of the application, but this should work for now.

### DIFF
--- a/src/components/card/utils.tsx
+++ b/src/components/card/utils.tsx
@@ -32,10 +32,13 @@ export const hasAuthorLockup = (variant: CardVariant, category: PillCategory) =>
 
 const parseContentDate = (
     contentDate: string | [string, string],
-    updateDate?: string
+    updateDate?: string,
+    eventTimeFormatting?: boolean
 ) => {
     if (Array.isArray(contentDate)) {
         return formatDateRange(contentDate[0], contentDate[1]);
+    } else if (eventTimeFormatting) {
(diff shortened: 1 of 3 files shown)


justinmckenzie (approved):

justinmckenzie:
@max-melo does this also update the "In-Person" label on the cards in the Events page? Just checking because I only the formatting to the Industry Events Details page in my PR in fear of breaking filtering logic. So we will just need to include them here if so

max-melo:
> @max-melo does this also update the "In-Person" label on the cards in the Events page? Just checking because I only the formatting to the Industry Events Details page in my PR in fear of breaking filtering logic. So we will just need to include them here if so

yeah this is already in the codebase here https://github.com/mongodb/devcenter/blob/2bc8e754f932197b312fc06e9d4c73a2dfeb5486/src/components/card/utils.tsx#L106
unfortunate that we have to have the same logic 3 times in the app, but hopefully we'll be able to refactor this later

harikakotipalli (approved):

harikakotipalli:
@max-melo LGTM. There are few User Group Meetups where timing on Community Forums UI is not matching our Cards.But I checked the community API JSON response and the timing conversion is correct, it's not on us if community API is giving wrong data.

PR #414, opened 2023-01-17
[DEVHUB-1631]: Create Reusable onboarding component

justinmckenzie:
## Jira Ticket:

[DEVHUB-1631](https://jira.mongodb.org/browse/DEVHUB-1631)

## Description:

Merges together all Modals in the app to use `useModal` context so all displaying logic of modals is handled in one place. Begins the two new Personalization modals that will be added (long scroll and paginated). There is still some work to be done on them like tying in the new topic cards, adding event handlers to display the modals, POST/PUT'ing the selections to the backend, and unit tests


justinmckenzie:
### Developer Notes

When it comes time to use the new Personalization Modals, you can do something like this for the respective modals (notice the long scroll one needs an additional argument to pass needed props to the modal component)

```
const { openModal } = useModalContext();

<button onClick={() => openModal(<PaginatedPersonalizationModal />)}>
    Open Paginated Modal
</button>
<button onClick={() => openModal(<ScrollPersonalizationModal />, { hideCloseBtn: true })}>
     Open Scroll Modal
</button>
```

justinmckenzie:
This is ready for review, I will just not merge until we cut a release branch for Events. Can work on unit tests in the meantime or they can be done when the relevant integration stories are written

max-melo:
not sure if it's just me but it seems like the pagination dots aren't reacting to going to the next page via the button

Screen Shot 2023-01-25 at 11 26 39 AM

i would expect the last dot to be filled here but it's stuck on the first one

src/components/modal/personalization/paginated/personalization-paginated.modal.tsx
```diff
@@ -0,0 +1,125 @@
...
+import {
+    initializePersonalizationConfig,
+    submitPersonalizationSelections,
+} from '../utils';
+import { tagToTopic } from '../../../../utils/tag-to-topic';
+
+import styles from '../styles';
+
+const PaginatedPersonalizationModal = () => {
+    const paginationConfig = initializePersonalizationConfig();
```
max-melo, line 25: can this be completely moved outside of the render loop, since it doesn't depend on any props or state? if i'm reading the code correctly, this is going to cause a ~100 iteration loop to run every time the component re-renders (ie any time a user interacts with the modal). i think maybe `initializePersonalizationConfig` can just be an exported array that's imported in this file, or we can consider even converting it into a preval'd value.

src/styled/layout.tsx
```diff
@@ -8,7 +8,7 @@ export const globalStyles = (hasOverlay: boolean) => `
     body {
         font-family: ${theme.fonts['euclid-circular-a']};
         margin: 0;
-        overflow: ${hasOverlay ? 'hidden' : 'visible'};
+        overflow: ${hasOverlay ? 'hidden' : 'visible'} !important;
```
max-melo, line 11: what's the reasoning for adding `!important` here?

max-melo:
I don't know exactly why, but these changes seem to cause a full re-render whenever a modal is opened or closed, which adds some lag/lack of responsiveness.

Before when opening a modal:
Screen Shot 2023-01-25 at 12 11 45 PM
-6946e978cdc7.png)

After:
Screen Shot 2023-01-25 at 12 11 40 PM

This doesn't make much sense to me because context was being used before as well. Maybe not too much of an issue for now, but I thought I'd flag it in case we want to look into it in the future.

src/components/modal/personalization/paginated/personalization-paginated.modal.tsx
max-melo: same here as well https://github.com/mongodb/devcenter/pull/414/files#diff-e7bda44f51bf7eb9974a43ecda03021180955a551dd45f7a97efd110c72b4e10R24

justinmckenzie: > I don't know exactly why, but these changes seem to cause what's almost a full page re-render whenever a modal is opened or closed, which adds some lag/lack of responsiveness.
> 
> Before when opening a modal: Screen Shot 2023-01-25 at 12 11 45 PM
> 
> After: Screen Shot 2023-01-25 at 12 11 40 PM
> 
> This doesn't make much sense to me because context was being used before as well. Maybe not too much of an issue for now, but I thought I'd flag it in case we want to look into it in the future.
> 
> EDIT: Did some digging and it has to do with [this line](https://github.com/mongodb/devcenter/pull/414/files#diff-0b75a532e530ce749a966d77dffeea3593e67560e5894213143637ee1d84aef5R57). Honestly I'm probably missing something but I've looked at the modals/overlays with and without this overflow switching and I can't tell a difference. Can this just be removed and `globalStyles` converted to a string instead of a function?

You're seeing the difference in the perf probably because the existing modals didn't try to set overflow: hidden and would just let the window scroll while the modal was open. I wasn't told this was _necessarily_ a defect, I just think its bad UX so I added the overflow hidden when all modals are open, especially because both personalization modals will have inner scroll, so its going to cause some wonky behavior if the window also has scroll while the user is trying to only scroll the modal

justinmckenzie: can look into a way getting a way around this without calling functions in styled if its a big issue, but I do think its something we're going to need to have

max-melo: gotcha, as far as the global styling I think we can just figure it out later, not too big of a deal to have half a second lag only when opening/closing modals

max-melo (changes requested):
Everything LGTM except the `initializePersonalizationConfig` & the `!important` comment

max-melo (approved):

src/components/modal/personalization/scroll/personalization-scroll.modal.tsx
```diff
@@ -0,0 +1,111 @@
...
+                        >
+                            {tags.map(tag => {
+                                const isSelected = !!selections.find(
+                                    prevTags => prevTags.name === tag.name
+                                );
+                                // do not pass slug b/c it would create a hyperlink on the card
+                                const { icon, title } = tagToTopic({
+                                    ...tag,
+                                    slug: '',
+                                });
```
mcolella14, line 78: This is my bad. I should have just removed the `href` prop from the `ThirdPartyLogo` [here](https://github.com/mongodb/devcenter/blob/main/src/utils/tag-to-topic.tsx#L25). There's no reason for the icon to be an anchor. If you remove the `href` prop from being passed in that file, you should be able to just pass the whole tag and not have to worry about an anchor being rendered.

mcolella14 (approved):
Looks great!

mcolella14 (approved):

src/components/modal/personalization/scroll/personalization-scroll.modal.tsx
justinmckenzie: @mcolella14 i tried that out and it seems like the lighthouse tests were failing removing the `href` prop from `<ThirdPartyLogo />`. Not sure why that would be the case, reverted for now but will try to find out whats causing that

mcolella14: Ah damn that's annoying, looks like flora still renders them as anchors even though they have no href https://storage.googleapis.com/lighthouse-infrastructure.appspot.com/reports/1675196997595-248.report.html#crawlable-anchors

justinmckenzie: @mcolella14 lol figured it was going to be something like that, ill look into their component and see if just conditionally setting the prop will alleviate it

mcolella14: ahh ok, that's fine with me

PR #417, opened 2023-01-18
[DEVHUB-1650] & [DEVHUB-1651]: Industry Event Details Page Fixes

justinmckenzie:
## Jira Ticket:

[DEVHUB-1650](https://jira.mongodb.org/browse/DEVHUB-1650)
[DEVHUB-1651](https://jira.mongodb.org/browse/DEVHUB-1651)

## Description:

Fixes date and time formatting and wrong CTA at bottom of Industry Events detail page

### DIFF
--- a/src/components/event-widget/event-widget.tsx
+++ b/src/components/event-widget/event-widget.tsx
@@ -7,6 +7,8 @@ import {
 } from '@mdb/flora';
 import { ThemeUICSSObject } from 'theme-ui';
 
+import { formatSingleDate } from '../../utils/format-date';
+
 import styles from './styles';
 
 interface EventWidgetProps {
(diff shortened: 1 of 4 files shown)


src/components/event-widget/event-widget.tsx
```diff
@@ -19,6 +19,20 @@ interface EventWidgetProps {
     wrapperStyles?: ThemeUICSSObject;
 }
 
+const formatDate = (d: string) => {
```
harikakotipalli, line 22: don't we have this function in @max-melo code already.

harikakotipalli (approved):

src/components/event-widget/event-widget.tsx
max-melo: i have a function for displaying a range of dates, but it can be used to format individual dates if the 2nd date in the range is passed as undefined. 

it may be worth having a util function that formats one date specifically to make the code less confusing, but i think it might be worth moving this to `src/utils/format-date.ts`

mcolella14 (approved):

src/components/event-widget/event-widget.tsx
harikakotipalli: agree with @max-melo on this. it might be worth moving this to src/utils/format-date.ts

justinmckenzie: sounds good, I tried passing a single date to Max's `formatDateRange`, but it had some formatting I don't need on the Industry Details page, so I'm just going to include this function in `utils/format-date.ts` in fear of breaking his functionality elsewhere by changing it.

PR #419, opened 2023-01-19
[DEVHUB-1653]: Industry Events Page Shows Parent Events With Child Events In Tertiary Nav

justinmckenzie:
## Jira Ticket:

[DEVHUB-1653](https://jira.mongodb.org/browse/DEVHUB-1653)

## Description:
- Fixes duplicate "Event" titles in Tertiary Nav when an Event has no tags (the child title produces a `events/events` slug which leads to a 404)

### DIFF
--- a/src/page-templates/content-page/content-page-template.tsx
+++ b/src/page-templates/content-page/content-page-template.tsx
@@ -458,7 +458,7 @@ const ContentPageTemplate: NextPage<ContentPageProps> = ({
                                 src={getPlaceHolderImage(image?.url)}
                                 sx={{
                                     borderRadius: 'inc30',
-                                    objectFit: 'fill',
+                                    objectFit: 'cover',
                                 }}
                                 layout="fill"
                             />
--- a/src/page-templates/content-page/content-page-template.tsx
+++ b/src/page-templates/content-page/content-page-template.tsx
@@ -458,7 +458,7 @@ const ContentPageTemplate: NextPage<ContentPageProps> = ({
                                 src={getPlaceHolderImage(image?.url)}
                                 sx={{
                                     borderRadius: 'inc30',
-                                    objectFit: 'cover',
+                                    objectFit: 'fill',
                                 }}
                                 layout="fill"
                             />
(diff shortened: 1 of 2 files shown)


max-melo (approved):

src/page-templates/content-page/content-page-template.tsx
```diff
@@ -458,7 +458,7 @@ const ContentPageTemplate: NextPage<ContentPageProps> = ({
                                 src={getPlaceHolderImage(image?.url)}
                                 sx={{
                                     borderRadius: 'inc30',
-                                    objectFit: 'cover',
+                                    objectFit: 'fill',
```
mcolella14, line 461: I'd say maybe just do this for events for now. This is having some stretching effects in existing articles 
Screen Shot 2023-01-19 at 12 34 18 PM
Screen Shot 2023-01-19 at 12 34 44 PM

mcolella14 (changes requested):

src/page-templates/content-page/content-page-template.tsx
justinmckenzie: @mcolella14  ah damn, checked out a bunch of different content types but didn't see these. FWIW though we need to solve for this outside of events. A happy medium I found was to use `objectFit: contain`, it won't fill the containers width if the image isn't that wide though (see screenshots), which I assume Mike will not be stoked about. Will keep looking to see if I can figure out a quick simple solution

mcolella14: Ahh ok yeah I think that works for me, but yeah unless we get the content writers to size their images correctly i think something is always gonna be off, but I think `contain` is a good medium for now!

mcolella14 (approved):

src/page-templates/content-page/content-page-template.tsx
justinmckenzie: @mcolella14 reverting the change after speaking with @harikakotipalli. She brought up the point that the user should adjust the image asset instead of us making a code change, since this hasn't been brought up in the last ~8 months of being live.

harikakotipalli (approved):

PR #422, opened 2023-01-20
[DEVHUB-1606] Follow and unfollow topics - Frontend

amybguo:
## Jira Ticket:

Link to jira ticket [DEVHUB-1606](https://jira.mongodb.org/browse/DEVHUB-1606)

## Description:

Frontend changes made to Topic, Content, and Topic x Content pages to allow users to follow and unfollow topics. 

-   [ ] I have performed a self-review(QA) of my code and reviewed with Design or Product (If Applicable)?

Topic page
Screen Shot 2023-01-20 at 1 59 46 PM (2)

Topic page - hover follow icon
Screen Shot 2023-01-20 at 1 59 32 PM

Topic page - just followed tooltip
Screen Shot 2023-01-20 at 1 59 35 PM (2)

Topic page - just unfollowed tooltip
Screen Shot 2023-01-20 at 1 59 38 PM (2)

Content page
Screen Shot 2023-01-20 at 2 01 14 PM

Content page - hover follow icon
Screen Shot 2023-01-20 at 2 01 16 PM (2)

Content page - hover unfollow icon
Screen Shot 2023-01-20 at 2 01 24 PM (2)


mcolella14 (commented):
Looks great! I think the Link and tooltip could be bundled together in a component, but it's not super necessary, we can refactor that later if we need to. I think this'll be good to go once we hook up the sign-in flows.

src/components/follow-link/follow-link.test.tsx
```diff
@@ -0,0 +1,185 @@
...
+import '@testing-library/jest-dom';
+
+import FollowLink from '.';
+import { Tag } from '../../interfaces/tag';
+
+import * as nextAuth from 'next-auth/react';
+import * as nextRouter from 'next/router';
+
+// @ts-expect-error hacky way to mock named exports
+nextRouter.useRouter = jest.fn().mockImplementation(() => ({ asPath: '' }));
```
max-melo, line 11: can this be accomplished via
```
const mockUseRouter = jest.fn(() => ({ asPath: '' }))
jest.mock('next/router', () => ({
    __esModule: true,
    ...jest.requireActual('next/router'),
    useRouter: mockUseRouter
}))
```
? might remove the need for a `// @ts-expect-error`

src/components/follow-link/follow-link.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+    // If we're not in a hoverTooltipText scenario, there should be no hover tooltip rendered.
+    const hasHoverTooltip = !!hoverTooltipText;
+
+    const onFollowClick = useCallback(() => {
+        let topics: Tag[];
+        // Never show the click tooltip when using the iconsOnly variant
+        if (!isFollowingAnyTopics) {
+            // TODO: Enter modal flow
+            alert('Onboarding modal will go here');
```
justinmckenzie, line 58: if you're interested in at least incorporating the modal to show in scope of this, you can do the below now that https://github.com/mongodb/devcenter/pull/414 is added in main.

```
      openModal(
        <ScrollPersonalizationModal
          title={<>You&apos;re now following this topic. <br /> Interested in following other topics?</>}
           existingSelections={[...I think you'll want to add the previously followed topic in the prop here so it's already pre-selected in the tags]}
        />,
        { hideCloseBtn: true }
```

justinmckenzie: but integration with the backend in the PUT's is not completed yet just FYI

src/components/follow-link/follow-link.tsx
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        if (followedTopics && isFollowing) {
+            topics = followedTopics.filter(
+                followedTopic => followedTopic.name !== topic.name
+            );
+        } else {
+            topics = followedTopics ? [...followedTopics, topic] : [topic];
+        }
+        console.log(topics); // TODO: POST followedTopics to external API endpoint.
```
justinmckenzie, line 78: the modal will handle the selected topics and posting to the backend, assuming all of the PUT bodies will look the same

src/utils/get-sign-in-url.ts
```diff
@@ -0,0 +1,14 @@
+import getConfig from 'next/config';
+
+const getSignInURL = (path: string): string => {
```
justinmckenzie, line 3: if it makes sense to, can you also incorporate this into `src/components/layout.tsx`? If it's different logic then ignore

src/components/follow-link/follow-link.tsx
mcolella14: Planning to add that into this PR

src/components/follow-link/follow-link.test.tsx
mcolella14: Ah yeah that seems a lot better, will do this

src/utils/get-sign-in-url.ts
mcolella14: ah yeah I think I meant to do that and forgot, thanks for pointing it out!

src/components/follow-link/follow-link.test.tsx
mcolella14: @max-melo I fixed this for all the instances where it mocks for the whole test file, but I couldn't get it to work with different implementations for each test, so I left those as is. If you know a good way to do this for multiple implementations let me know. I tried `doMock`, which is [what the docs seem to suggest](https://jestjs.io/docs/jest-object#jestdomockmodulename-factory-options), but it didn't have the desired behavior.

max-melo: If I remember correctly you should be able to pass in a variable into the mock and then use `mockImplementationOnce` in each test. Just have to make sure that when mocking, you're returning a function that calls the mock instead of the mock itself or you'll get errors (something I forgot in my original comment)

```
const mockUseSession = jest.fn(())
jest.mock('next-auth/react', () => ({
    __esModule: true,
    ...jest.requireActual('next-auth/react'),
    useSession: (...args) => mockUseSession(...args)
}))
...
test('...', () => {
    mockUseSession.mockImplementationOnce(() => ({ ... }))
    ...
})
```
I think this should work, but if not I'm fine with leaving the original implementation

mcolella14: Ahhh ok yep I tried the exact thing you're talking about, tried to return the mock function itself: 
```useSession: mockUseSession```

 I have it how you have it now and it seems to be working, thank you!

max-melo (approved):

justinmckenzie (approved):

PR #424, opened 2023-01-23
Devhub 1617 get user preferences and persist user preferences in session

harikakotipalli:
[DEVHUB-1617](https://jira.mongodb.org/browse/DEVHUB-1617) - get user preferences and persist user preferences

### DIFF
--- a/next-auth.d.ts
+++ b/next-auth.d.ts
@@ -1,4 +1,5 @@
 import { DefaultSession } from 'next-auth';
+import { Tag } from './src/interfaces/tag';
 
 declare module 'next-auth' {
     /**
@@ -14,6 +15,10 @@ declare module 'next-auth' {
         lastName: string;
         /** Custom: The user's email address, taken from profile. */
         email: string;
+        /** additional fields added as part of personalization **/
+        followedTags?: Tag[] | null;
+        lastLogin: string | null;
+        emailPreference: boolean;
     }
 
     interface Profile {
--- a/next-auth.d.ts
+++ b/next-auth.d.ts
@@ -1,6 +1,4 @@
-import NextAuth, { DefaultSession } from 'next-auth';
-import { Session } from 'next-auth';
-import { JWT } from 'next-auth/jwt';
+import { DefaultSession } from 'next-auth';
 
 declare module 'next-auth' {
     /**
@@ -35,5 +33,6 @@ declare module 'next-auth/jwt' {
         lastName: string;
         /** Custom: The user's email address, taken from profile. */
         email: string;
+        userId?: string;
     }
 }
--- a/src/interfaces/UserPreference.ts
+++ b/src/interfaces/UserPreference.ts
@@ -1,0 +1,7 @@
+import { TagType } from '../types/tag-type';
+
+export interface UserPreference {
+    tag_name: string;
+    category: TagType;
+    slug: string;
+}
--- a/src/interfaces/UserPreference.ts
+++ b/src/interfaces/UserPreference.ts
@@ -1,7 +1,0 @@
-import { TagType } from '../types/tag-type';
-
-export interface UserPreference {
-    tag_name: string;
-    category: TagType;
-    slug: string;
-}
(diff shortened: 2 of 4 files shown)


src/interfaces/UserPreference.ts
```diff
@@ -0,0 +1,7 @@
+import { TagType } from '../types/tag-type';
+
+export interface UserPreference {
+    tag_name: string;
```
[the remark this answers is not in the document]
harikakotipalli: https://docs.google.com/document/d/15-M7pJnPamnlUbLjoy0rul7tlWEAGNkHbOG3k0_ttEo/edit - following same convention as here for preference.

src/pages/api/auth/[...nextauth].ts
```diff
@@ -2,6 +2,46 @@ import OktaProvider from 'next-auth/providers/okta';
 import type { NextAuthOptions } from 'next-auth';
 import type { NextApiRequest, NextApiResponse } from 'next';
 import NextAuth from 'next-auth';
+import * as Sentry from '@sentry/nextjs';
+
+async function getUser(userId: string): User | null {
+    const query = 'userId=' + userId;
+    const url = `${process.env.PERSONALIZATION_URL}/user_preferences?${query}`;
+    const options = {
+        method: 'GET',
```
Comment (mcolella14) [inline src/pages/api/auth/[...nextauth].ts:12]: `GET` is the default method so I don't think it needs to be explicitly specified here.

src/pages/api/auth/[...nextauth].ts
```diff
@@ -26,12 +66,35 @@ export const nextAuthOptions: NextAuthOptions = {
...
             session.email = token.email;
+            session.userId = token.userId;
+            const user = await getUser(token.userId);
+            if (user == null) {
+                const newUser = {
+                    user_id: token.userId,
+                    first_name: token.firstName,
+                    last_name: token.lastName,
+                    email: token.email,
+                    preferences: [],
```
Comment (mcolella14) [inline src/pages/api/auth/[...nextauth].ts]: Kinda goes along with the interface discussion, but could we call this `followed_topics` or something? I feel like we might have more preferences that the user can set in the future, and it could get confusing if the `preferences` field is just the topics they follow.

src/pages/api/auth/[...nextauth].ts
```diff
@@ -2,6 +2,46 @@ import OktaProvider from 'next-auth/providers/okta';
...
+
+async function getUser(userId: string): User | null {
+    const query = 'userId=' + userId;
+    const url = `${process.env.PERSONALIZATION_URL}/user_preferences?${query}`;
+    const options = {
+        method: 'GET',
+        headers: {
+            Accept: 'application/json',
+            'Content-Type': 'application/json',
+            apiKey: process.env.REALM_API_KEY,
```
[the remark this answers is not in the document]
Comment (harikakotipalli) [reply src/pages/api/auth/[...nextauth].ts]: There is not much logic in the http function for this , I will add one in fast api too , but the suggestion from Deepak is to go with server less initially.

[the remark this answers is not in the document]
Comment (mcolella14) [reply src/pages/api/auth/[...nextauth].ts]: I would definitely prefer we use our backend for this and everything else in personalization. We kinda decided against using serverless functions for a variety of reasons, and the goal is to move everything off of Realm.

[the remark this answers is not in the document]
Comment (harikakotipalli) [reply src/pages/api/auth/[...nextauth].ts]: Sure thing. I'm opening a PR shortly for backend and we can switch the API's to fast api, until then front end can use these.

next-auth.d.ts
```diff
@@ -1,6 +1,4 @@
-import NextAuth, { DefaultSession } from 'next-auth';
-import { Session } from 'next-auth';
-import { JWT } from 'next-auth/jwt';
+import { DefaultSession } from 'next-auth';
 
 declare module 'next-auth' {
```
mcolella14, line 3: Should `followedTags`, `lastLogin`, and `emailPreference` be added to the `Session` interface here since we add those values?

justinmckenzie: @mcolella14 @harikakotipalli if we do this then we can likely just use the `Session` interface instead of having to create the `User` interface too right?

harikakotipalli: @justinmckenzie Not all fields in user need to be in session - what if there is some client side call eg: home page content , which deals with new field which needs to be in User interface , but not needed in session.

harikakotipalli: @mcolella14 : +1 , adding those fields to session. If I don't add I don't think we can access them in useSession, getSession, <SessionProvider> etc.

mcolella14 (approved):

PR #429, opened 2023-01-26
Fix Events Results not showing when JavaScript is disabled

max-melo:
For SEO reasons, results should be generated statically in the build output, as well as dynamically, which can be verified by disabling javascript. Previously, doing so resulted in a "No Results" screen on the Events page, but this PR should fix that issue. I also created a utility function `isServerSide()` instead of checking `typeof window === 'undefined'` every time.

### DIFF
--- a/src/components/card/card.tsx
+++ b/src/components/card/card.tsx
@@ -33,6 +33,7 @@ import { getURLPath } from '../../utils/format-url-path';
 import { h5Styles, h6Styles } from '../../styled/layout';
 import EventIcon from '../icons/event-icon';
 import { PillCategory } from '../../types/pill-category';
+import isServerSide from '../../utils/is-server-side';
 
 const CardThumbnail = ({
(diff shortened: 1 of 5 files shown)


justinmckenzie (approved):

mcolella14 (approved):

gfunk1230 (approved):
LGTM

max-melo:
Seeing some strange playwright-related errors for the events page that I can't reproduce on my local--they're related to visual testing (the visual test is timing out in GitHub actions waiting for network idle) and Lighthouse is complaining about `robots.txt`. I've verified that the site does indeed still have a `robots.txt` and that no network requests are idling. 

These seem to be solely related to these tests in the GitHub Actions environment, so I'm going to merge this for now since we're on a tight schedule. However, I'm going to continue looking into these test failures and hopefully figure out what's going on.

PR #446, opened 2023-02-10
[part of DEVHUB-1688] Mock Service Workers POC

max-melo:
## Jira Ticket:

[DEVHUB-1688](https://jira.mongodb.org/browse/DEVHUB-1688)

## Description:

- Integrate MSW as the main API mocking strategy. 
- Add mocked run scripts to package.json
- Add an `update-mocks` script that fetches from all of the endpoints and updates the local mocks. Can be run via `yarn run update-mocks`

Had a lot of issues trying to get this to work with `next-plugin-preval`. Initially wanted/tried to use a webpack loader, but it turns out that `next-plugin-preval` [requires the file directly](https://github.com/ricokahler/next-plugin-preval/blob/5883342d87e1d64b29c20af8732f92f6b94f77da/src/loader.ts#L104) and is unaffected by previous loaders. Because of this, the only way that I've found to get the mocking to work is to import MSW in each preval file.

Also, I'm using the full API response for all of the endpoints here so the site fully functions in mocked mode. This means that the mocks folder is pretty large, around 36 MB:
Screen Shot 2023-02-10 at 8 55 19 AM

I can look into ways of paring this down if this is an issue.

-   [x]  I have commented my code, particularly in hard-to-understand areas

Here's the app running via `yarn dev:mocked`, where asterisks indicate that that piece of content is mocked (asterisks can be removed if needed, but without them it's hard to tell that you're looking at mocked content):
Screen Shot 2023-02-10 at 8 52 24 AM
Here's some logging output indicating that calls to the respective endpoints are being mocked with local JSON
Screen Shot 2023-02-10 at 8 52 07 AM


package.json
```diff
@@ -3,16 +3,20 @@
...
         "start": "next start",
+        "start:mocked": "NEXT_PUBLIC_API_MOCKING=enabled next start",
         "lint": "next lint",
         "format:write": "prettier --write .",
         "format:check": "prettier --check .",
         "test": "jest --watch",
         "test:coverage": "jest --coverage",
-        "test:e2e": "npx playwright install --with-deps && npx playwright test",
-        "test:e2e:visual": "npx @percy/cli exec -- npm run test:e2e"
+        "test:e2e": "yarn build:mocked && npx playwright install --with-deps && npx playwright test",
```
mcolella14, line 17: Sometimes when testing locally, I'll have the app running in development mode, and then run playwright tests on that to see changes in real time. With this change, I'll have to wait for the app to build before the tests run. Would it be possible to just add back that build step in the GitHub Actions pipeline and change it to`yarn build:mocked` instead?

mcolella14 (commented):
Looks great! Just the one nitpick.

On the topic of the size of the `mocks` folder, I think it'll be fine as long as you add an entry to the `.dockerignore`file. This way those won't be copied to the images in staging/production.

mcolella14 (approved):

justinmckenzie (approved):

PR #447, opened 2023-02-14
[DEVHUB-1606] Fix backend compatibility, fix follow-link functionality.

mcolella14:
## Jira Ticket:

Partially related to [DEVHUB-1606](https://jira.mongodb.org/browse/DEVHUB-1606).

## Description:

Implements the backend connectivity for the FollowLink component. There were some weird things about this:
* Added a hacky way to refresh the session manually. It is imperfect though since it just emits an event on the browser and we can't await the refresh.
* Added an extra state `isShowingFollowing` to manually control what the link displays. This was necessary because after we make the API call to update the user, we trigger an event to refresh the session. However, we have no way of knowing when the new request to get the session will complete, since it is just triggered by an event we emit. In order to avoid sync issues between the button text and the tooltip text, I created this state to provide a clean UI experience. This state is reset to the session value in the `useEffect`, so even if it happens to be out of sync with the actual session, it will fall back to the correct value eventually.
* Needed to add some debouncing logic, but had to change the original function since it was being reevaluated every time we called it since it toggles `isShowingFollowing`.

### DIFF
--- a/src/components/follow-link/follow-link.tsx
+++ b/src/components/follow-link/follow-link.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useMemo, useState } from 'react';
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
 import debounce from 'lodash.debounce';
 
 import { Link, SystemIcon, ESystemIconNames } from '@mdb/flora';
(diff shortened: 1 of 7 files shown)


src/utils/refresh-session.ts
```diff
@@ -0,0 +1,8 @@
+// See  https://github.com/nextauthjs/next-auth/issues/596#issuecomment-943453568
+// Next auth will refetch the session on this event, so we manually trigger it.
+
+export default function refreshSession() {
+    console.log('Displathci');
```
justinmckenzie, line 5: console

src/components/follow-link/follow-link.tsx
```diff
@@ -48,53 +52,73 @@ const FollowLink: React.FunctionComponent<FollowLinkProps> = ({
...
-                followedTopic => followedTopic.name !== topic.name
-            );
-        } else {
-            topics = followedTopics ? [...followedTopics, topic] : [topic];
-        }
-        console.log(topics); // TODO: POST followedTopics to external API endpoint.
-    }, [isFollowing, followedTopics, topic]);
+            submitPersonalizationSelections({
+                followedTags,
+                emailPreference: session.emailPreference,
```
justinmckenzie, line 104: 👍 nice catch here, i woulda just though to pass `false` lol

justinmckenzie (approved):

max-melo (approved):

PR #450, opened 2023-02-21
Fix for Recommended Content Modal

max-melo:
Quick fix/follow up to https://github.com/mongodb/devcenter/pull/448
Bringing up the personalization modal when a user has followed topics would always display no tags selected. Now, the tags initially selected should reflect what the user is currently following.

edit: also fixed caching behavior that was noticed with recommended content endpoint

### DIFF
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -82,10 +82,9 @@ const Home: React.FunctionComponent<HomeProps & NextPage> = ({
     recommendedTags,
 }) => {
     const { status, data } = useSession();
-    const { followedTags } = data || {};
-    const { data: content, isValidating } = usePersonalizedContent(
-        followedTags || []
-    );
+    const followedTags = data?.followedTags || [];
+    const { data: content, isValidating } =
+        usePersonalizedContent(followedTags);
 
     return (
         <main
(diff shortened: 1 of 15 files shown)


src/pages/index.tsx
```diff
@@ -155,13 +154,13 @@ const Home: React.FunctionComponent<HomeProps & NextPage> = ({
...
                     tags={recommendedTags}
+                    followedTags={followedTags}
                     content={content}
                     showFooter
                     onTagsSaved={(
                         followedTags: Tag[],
                         emailPreference: boolean
                     ) => {
-                        // TODO: Replace with call to user preferences endpoint
                         submitPersonalizationSelections({
```
justinmckenzie, line 164: if you want/have the capacity, you can add the below so that the Toasts can appear for success/failure when saving the recommended tags (it's already added on the modal callbacks so no need to add it there). I have an active branch for a couple small fixes so I can add it there if needed.

```
        const { error } = await submitPersonalizationSelections({
                 followedTags,
                 emailPreference,
        });

        if (!error) {
            setNotification({
                message: 'Successfully saved your preferences',
                variant: 'SUCCESS',
            });
        } else {
            setNotification({
                message:
                    'Your request could not be completed at this time. Please try again.',
                variant: 'WARN',
            });
        }
```

justinmckenzie (approved):

harikakotipalli (approved):

mcolella14 (approved):

PR #452, opened 2023-02-22
[DEVHUB-1703] Implement remaining notifications.

mcolella14:
## Jira Ticket:

[DEVHUB-1703](https://jira.mongodb.org/browse/DEVHUB-1703)

## Description:

- Added a hook (`useUserPreferences`) to utilize the notification context whenever updating preferences.
- Added `failedToFetch` flag on the session object to indicate that we failed to find the user's information.
- Adjust `submitPersonalizationSelections` to throw errors consistently.
- Used these to implement notifications.
- Also adjusted the `Cache-Control` header on personalization endpoints to invalidate the browser cache when sending a bad response. I was having some issues with this and the SWR cache, but it seems to work now.

### DIFF
--- a/src/hooks/personalization/user-preferences.ts
+++ b/src/hooks/personalization/user-preferences.ts
@@ -1,6 +1,7 @@
 import { useCallback } from 'react';
 import { submitPersonalizationSelections } from '../../components/modal/personalization/utils';
 import { Tag } from '../../interfaces/tag';
+import { useNotificationContext } from '../../contexts/notification';
 
 interface UserPreferences {
     followedTags: Array<Tag>;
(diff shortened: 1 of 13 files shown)


justinmckenzie (approved):
nice thinking putting everything in a hook

src/hooks/personalization/user-preferences.ts
```diff
@@ -0,0 +1,45 @@
...
+                    variant: 'WARN',
+                });
+                return;
+            }
+            setNotification({
+                message: 'Successfully saved your preferences',
+                variant: 'SUCCESS',
+            });
+        },
+        []
```
max-melo, line 37: is there a particular reason why `setNotification` isn't in the dependency array? if not, i think it's worth adding... not sure if react guarantees the context value is populated before every render, but i'm seeing a default `() => null` value for setNotification, so best to be safe imo

mcolella14: Ahh yep I just missed that, will add that, thanks for the catch!

harikakotipalli (approved):

max-melo (approved):
LGTM!

PR #471, opened 2023-03-14
[SKUNK] Optimizations - Upgrade to Next13, reduce bundle size, fix security bug, etc.

mcolella14:
Worked on some optimizations during skunkworks. Wanted to make the client-side first-load js bundles smaller, but flora is not optimized for that. Still, the total bundle size now does not include the mocks, and we are on Next.js 13, which has a number of unstable features to help performance that we can leverage once they become stable.

### DIFF
--- a/src/components/hero/hero.test.tsx
+++ b/src/components/hero/hero.test.tsx
@@ -22,6 +22,10 @@ jest.mock('../../utils/get-sign-in-url', () => {
     return jest.fn(() => '');
 });
 
+jest.mock('../modal/personalization', () => {
+    return jest.fn(() => <></>);
+});
+
 const crumbs = [
     { text: 'MongoDB Developer Center', url: '/' },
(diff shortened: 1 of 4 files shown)


src/components/expanding-link/expanding-link.test.tsx
```diff
@@ -5,38 +5,41 @@ import theme from '@mdb/flora/theme';
...
+    const user = userEvent.setup();
 
     const link = screen.getByRole('link');
 
     expect(link).toHaveStyle({
         right: 0,
     });
 
-    userEvent.hover(link);
+    await user.hover(link);
```
justinmckenzie, line 42: would using `act()` here fit your use case instead of `async/await` ? https://reactjs.org/docs/test-utils.html#act https://flyingsquirrel.medium.com/when-should-i-use-act-in-react-testing-library-d7dd22a3340e

PR #483, opened 2023-04-11
[DEVHUB-1753] Migrate Meta Info, Authors, and Articles to use Contentstack.

mcolella14:
## Jira Ticket:

[DEVHUB-1753](https://jira.mongodb.org/browse/DEVHUB-1753)
## Description:
Using the ContentStack GraphQL API here. This is pulling data from the "prod" alias in ContentStack.

The key differences are the pagination (there is a limit of 100 record per GraphQL call), and structure of the queries and responses. This makes mocking a bit more tough, but I believe the solution I have in place should work for most things left.

I also changed how the authors and articles are mapped. Authors no longer have their content stored on their own records, we just map all content to the authors by author slug.


src/interfaces/author.ts
```diff
@@ -1,27 +1,28 @@
...
+    image?: Image;
+    title?: string;
+    name: string;
+}
+
+export interface AuthorResponse extends AuthorBase {
+    imageConnection?: ImageResponse;
+    job_title?: string;
+    title: string;
+}
```
mcolella14, line 28: Had to separate these two interfaces, since the fields in CS are differently named than the fields in Strapi. Also, I've removed `articles` here because we don't really need it. All content items already reference an author, so we can just filter the content items preval to find all content by a particular author. See [here](https://github.com/mongodb/devcenter/pull/483/files#diff-c5a9b76b1e3eb322d6e1d7f6601467d4cf9fc6eff73249f5cf26dc2d5114ad1fR264-R269).

src/interfaces/meta-info.ts
```diff
@@ -11,21 +11,46 @@ export interface MetaInfo {
...
+    ['code_levels', 'CodeLevel'],
+    ['event_types', 'EventType'],
+    ['event_attendances', 'EventAttendance'],
+]);
 
-interface L1Props {
-    l_1_product?: L1Props2;
+interface L1ProductsConnection {
+    edges: { node: { title: string } }[];
 }
```
mcolella14, line 43: This is just how CS represents relations, so we have to conform to this structure.

PR #484, opened 2023-05-02
MongoDB TV

mcolella14:
## Jira Ticket:

[DEVHUB-1763](https://jira.mongodb.org/browse/DEVHUB-1763)
## Description:
Doc for reference: https://docs.google.com/document/d/1zxcU8aO46iXDeknC3qy1EPAJVbp4cNwT9N_bFOGmYQI

- Creates content pages for MongoDB TV content when appropriate.
- Adds new sorting mode for Videos and will surface MongoDB TV content accordingly.
- Handles data for MongoDB TV search items and renders cards with date ranges.
- Changes some branding of Videos to MongoDB TV/Shows.
- Adds a "Subscribe" CTA to the videos page.


src/api-requests/get-mongodb-tv-shows.ts
```diff
@@ -0,0 +1,48 @@
...
+        show => '/videos/' + slugify(show.title) === slug
+    );
+    if (!show) {
+        const e = new Error(
+            `Could not find MongoDBTV show with slug "${slug}"`
+        );
+        Sentry.captureException(e);
+        throw e;
+    }
+    return show;
```
justinmckenzie, line 47: is it ok to returned `undefined` here if `show` isn't found? i see you're assigning the return value to `content` in `/get-content-by-slug.tsx`. Not sure how `content` is used after that but wanted to flag just in case

mcolella14: Ah so it is supposed to return `null` if it's not found, I will update that.

justinmckenzie (approved):

harikakotipalli (approved):
@mcolella14  LGTM. Any reason why mock data is changed as part of this PR.

mcolella14:
@harikakotipalli Yeah so the reason for the mock changes is because I had to add mocking for the MongoDB TV content, and running `yarn build-mocks` just updates all mocks, so all are updated in the PR. I figure it's good to have updated data anyway.

src/service/build-content-items.ts
```diff
@@ -41,7 +44,63 @@ export const mapPodcastsToContentItems = (
...
+    const slug = text.replace(/[^a-zA-Z0-9]+/g, '-');
+    return slug.toLowerCase();
+};
+
+export const mapMongoDBTVShowsToContentItems = async (
+    allMongoDBTVShows: MongoDBTVShow[],
+    isRuntime: boolean
+) => {
+    let mappedMongodbTVShows: ContentItem[] = [];
+    // we can leverage preval if it's runtime.
```
harikakotipalli, line 68: @mcolella14 : what is the meaning of runtime ?

src/service/get-all-tags.ts
```diff
@@ -0,0 +1,47 @@
...
+    getAllTechnologiesMetaInfo,
+} from '../api-requests/get-all-meta-info';
+
+import {
+    parseMetaInfoResponse,
+    parseMetaInfoResponseForL1,
+} from './get-all-meta-info';
+
+// Similar to getAllMetaInfo, but doesn't filter out tags without content associated.
+export const getAllTags = async (): Promise<MetaInfo[]> => {
```
harikakotipalli, line 18: Will this be pointing to Contentstack later on ?

src/service/build-content-items.ts
mcolella14: As opposed to build time. If we call this method at build time, we cannot use preval data since it will not have been built, but if we call it while the app is running (like if a user makes a request and we generate a page server-side), we can use that preval data.

src/service/get-all-tags.ts
mcolella14: Yes it will. I imagine all the `get<CONTENT_TYPE>MetaInfo` methods will point to ContentStack anyway, so there should be much to change here.

PR #512, opened 2023-06-09
Cms migration

harikakotipalli:
Allotting CPU and Memory resources as suggested by Kanopy team

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -4,6 +4,11 @@ kind: pipeline
 type: kubernetes
 name: build-devcenter-staging
 
+resources:
+  requests:
+    cpu: 2000
+    memory: 18Gi
+
 steps:
   - name: build
     image: node:16.16.0-slim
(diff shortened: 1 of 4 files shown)


.drone.yml
```diff
@@ -4,6 +4,11 @@ kind: pipeline
 type: kubernetes
 name: build-devcenter-staging
 
+resources:
+  requests:
+    cpu: 2000
+    memory: 18Gi
```
jackyw2018, line 10: Should we write it as 2000m? 

"Limits and requests for CPU resources are measured in cpu units. In Kubernetes, 1 CPU unit is equivalent to 1 physical CPU core, or 1 virtual core, depending on whether the node is a physical host or a virtual machine running inside a physical machine.

Fractional requests are allowed. When you define a container with spec.containers[].resources.requests.cpu set to 0.5, you are requesting half as much CPU time compared to if you asked for 1.0 CPU. For CPU resource units, the [quantity](https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/) expression 0.1 is equivalent to the expression 100m, which can be read as "one hundred millicpu". Some people say "one hundred millicores", and this is understood to mean the same thing."

https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/

mcolella14 (approved):

PR #520, opened 2023-06-21
Contentstack preview

harikakotipalli:
(Matt speaking) The majority of the work came is located in the `pages/cs_preview` and `build-content-items` files. The gist is that Live Preview is only compatible with the ContentStack SDK, so for this preview route we have to use the SDK to fetch the data to populate the template. Unfortunately the response data structure is a bit different than that of the GraphQL API, so most of the work here is just extra functionality to parse that structure.

I also did some minor refactoring, but it's pretty straightforward.

Here are the resources I used to do this:

https://www.contentstack.com/docs/developers/set-up-live-preview/set-up-live-preview-for-your-website/

https://www.contentstack.com/docs/developers/set-up-live-preview/how-live-preview-works-with-sdk/


src/config/api-client.ts
```diff
@@ -21,14 +21,13 @@ const CS_HEADERS_STAGING = {
...
     'ApolloGraphQL',
     CS_URL_STAGING,
     CS_HEADERS_STAGING
 );
-export const CS_CLIENT_MOCK = clientFactory('Mock', CS_URL, CS_HEADERS);
+const useMock = process.env.NEXT_PUBLIC_API_MOCKING === 'enabled';
+
+const PROD_CLIENT_NAME = useMock ? 'Mock' : 'ApolloGraphQL';
+
+export const CS_CLIENT = clientFactory(PROD_CLIENT_NAME, CS_URL, CS_HEADERS);
```
mcolella14, line 33: Prevents us from unnecessarily calling `clientFactory('Mock', ...)` when we are not mocking.

jackyw2018: Love this refactoring!

src/utils/stack.js
```diff
@@ -0,0 +1,14 @@
+import Contentstack from 'contentstack';
+
```
jackyw2018, line 2: Although I just re-factored the gql queries using apolloClient, we may want to refactor it again using CS-Delivery SDK to query data now that we have contentstack installed...

jackyw2018: Sigh should have thought about this... but here is the [ticket](https://jira.mongodb.org/browse/DEVHUB-1831)

jackyw2018 (approved):
LGTM

PR #528, opened 2023-06-28
[DEVHUB]-1807 Stale Responses from ContentStack Graphql API's impacting previews

jackyw2018:
## Jira Ticket:

[DEVHUB-1807](https://jira.mongodb.org/browse/DEVHUB-1807)

## Updates:

1. Updated `fetchAll()` to have a `fetchPolicy`; for staging-related requests, `no-cache` is used to avoid stale responses
2. Changed ApolloClient from GET (with URI encoding) to POST (default option) to avoid occasional URI-too-long error code from Contentstack
3. Updated UI to render 404 page (instead of a 5xx error) when a content is not found, which happens when a preview content is not found

### DIFF
--- a/src/pages/preview/[...slug].tsx
+++ b/src/pages/preview/[...slug].tsx
@@ -9,11 +9,12 @@ import {
     getPreviewContentForArticles,
     getPreviewContentForEvents,
 } from '../../service/get-preview-content';
+import Error from 'next/error';
 
 interface ContentPageProps {
     crumbs: Crumb[];
     topic: Tag;
-    contentItem: ContentItem;
+    contentItem: ContentItem | null;
     relatedContent: ContentItem[];
     tertiaryNavItems: TertiaryNavItem[];
     previewMode?: boolean;
@@ -27,6 +28,10 @@ const ContentPage: NextPage<ContentPageProps> = ({
     relatedContent,
     previewMode,
 }) => {
+    if (!contentItem) {
+        return <Error statusCode={404} />;
+    }
+
     return (
         <ContentPageTemplate
             crumbs={crumbs}
@@ -49,7 +54,7 @@ export const getServerSideProps = async (context: any) => {
         type: 'Technology',
         slug: '',
     };
-    let contentItem;
+    let contentItem: ContentItem | null;
     if (slugString.startsWith('events/')) {
         contentItem = await getPreviewContentForEvents('/' + slugString);
     } else {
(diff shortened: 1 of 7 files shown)


src/pages/preview/[...slug].tsx
```diff
@@ -27,6 +28,10 @@ const ContentPage: NextPage<ContentPageProps> = ({
     relatedContent,
     previewMode,
 }) => {
+    if (!contentItem) {
+        return <Error statusCode={404} />;
+    }
+
```
Comment (mcolella14) [inline src/pages/preview/[...slug].tsx:34]: This is kinda unrelated, but Patrick mentioned that it might be better for us to only have preview routes accessible on staging to prevent them from technically being public. Do you think we should just also return a 404 here if we are on prod?

Comment (jackyw2018) [reply src/pages/preview/[...slug].tsx:34]: Yeah, fair; I did find it risky to expose preview even though people need the correct URL to access;

I can update this, but do we need to check with Greg?

Comment (mcolella14) [reply src/pages/preview/[...slug].tsx:34]: Yeah i think it would be best to loop him in

Comment (gfunk1230) [reply src/pages/preview/[...slug].tsx:34]: You mention the authors can potentially experience breaking changes on the staging app, but I would like to know if that is something we are encountering now or have in the past. Or are we seeing potential in that happening

Comment (jackyw2018) [reply src/pages/preview/[...slug].tsx:34]: It seems like the consensus is to keep the preview exclusive to the staging app. I will update it accordingly!

Comment (jackyw2018) [reply src/pages/preview/[...slug].tsx:34]: > You mention the authors can potentially experience breaking changes on the staging app, but I would like to know if that is something we are encountering now or have in the past.

Not really. I just wrote that to be precautions, as the staging app (i.e., main branch) should largely be stable anyway.

src/pages/preview/[...slug].tsx
```diff
@@ -27,6 +28,9 @@ const ContentPage: NextPage<ContentPageProps> = ({
     relatedContent,
     previewMode,
 }) => {
+    const isStaging = process.env.APP_ENV === 'staging';
+    if (!contentItem || isStaging) return <Error statusCode={404} />;
```
Comment (mcolella14) [inline src/pages/preview/[...slug].tsx:32]: I think it might be best to do this in `getServerSideProps` just to ensure that we return a 404 from the server. Here is an example of us doing that in the regular [...slug] page https://github.com/mongodb/devcenter/blob/main/src/pages/%5B...slug%5D.tsx#L106

PR #538, opened 2023-07-06
[DEVHUB-1824] Filter out past events in UI

jackyw2018:
## Jira Ticket:

[DEVHUB-1824](https://jira.mongodb.org/browse/DEVHUB-1824)

## Updates:

- Add filtering logic in frontend to filter out past events

### DIFF
--- a/src/page-templates/content-type-page/content-type-data.ts
+++ b/src/page-templates/content-type-page/content-type-data.ts
@@ -40,6 +40,12 @@ export const getContentTypePageData = async (
         );
     }
 
+    if (initialSearchContent && contentType === 'Event') {
+        initialSearchContent = initialSearchContent.filter(
+            item => item.end_time && new Date(item.end_time) > currentTime
+        );
+    }
+
     const filterItems = await getFilters(contentType, allSearchContent);
 
     const metaInfoForTopic = await getMetaInfoForTopic(slug);


src/page-templates/content-type-page/content-type-data.ts
```diff
@@ -40,6 +40,12 @@ export const getContentTypePageData = async (
         );
     }
 
+    if (initialSearchContent && contentType === 'Event') {
+        initialSearchContent = initialSearchContent.filter(
+            item => item.end_time && new Date(item.end_time) > currentTime
```
jackyw2018, line 45: Do we have plans to migrate the realm search function into our backend?

Also, should we expand the search function with a time param (and potentially more params for other use cases) to apply filtering logic at the backend and further reduce UI processing?

jackyw2018: Also let me know if this filtering logic should happen elsewhere in case this is not the right place.

mcolella14: Yes there are plans for this, in this ticket here https://jira.mongodb.org/secure/RapidBoard.jspa?rapidView=2027&projectKey=DEVHUB&view=detail&selectedIssue=DEVHUB-1738&quickFilter=16572. There are just some steps that need to be taken before actually migrating.

And so I think eventually once it's moved to the backend, we can just do the date compare on the backend and handle the filtering there, no need to pass the time as a parameter.

mcolella14:
Actually I think you will want to add something similar to what MongoDB TV is doing here: https://github.com/mongodb/devcenter/blob/main/src/hooks/search/index.ts#L170-L183. 

As it is right now, I think the initial search results when the page is loaded will be correct, but if someone uses a search query, past events will still populate.

mcolella14 (approved):

PR #544, opened 2023-07-11
Add more SEO fields

mcolella14:
## Description:

- Provide a default `og:type` of `article` or `video:other` when appropriate.
- Add missing GraphQL twitter-related fields.
- Add `publishedTime`, `modifiedTime`, and `authors` for articles.

### DIFF
--- a/src/page-templates/content-page/content-page-template.tsx
+++ b/src/page-templates/content-page/content-page-template.tsx
@@ -619,6 +619,13 @@ const ContentPageTemplate: NextPage<ContentPageProps> = ({
             </>
         );
     };
+    const ogType =
+        seo?.og_type ||
+        ['Article', 'Code Example', 'Quickstart', 'Tutorial'].includes(category)
+            ? 'article'
+            : category === 'Video'
+            ? 'video:other'
+            : undefined;
 
     return (
         <>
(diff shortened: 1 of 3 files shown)


src/page-templates/content-page/content-page-template.tsx
```diff
@@ -632,11 +639,18 @@ const ContentPageTemplate: NextPage<ContentPageProps> = ({
...
                     description: seo?.og_description,
                     images: seo?.og_image?.url
                         ? [{ url: seo?.og_image?.url }]
                         : [],
+                    article: {
+                        publishedTime: Array.isArray(contentDate)
+                            ? undefined
+                            : contentDate,
+                        modifiedTime: updateDate,
+                        authors: authors.map(({ name }) => name),
```
jackyw2018, line 652: What should the content of authors field be? I have come across different examples:

1. The example in the [documentation](https://github.com/garmeeh/next-seo) is a list of URLs.
2. LinkedIn Open graph shows a [profile array](https://www.linkedin.com/pulse/open-graph-protocol-atoconn-india/)

mcolella14: Ah good call, I just saw an `string[]` in the documentation and assumed it would be names. I can replace that with a URL pointing to the author page.

jackyw2018:
LGTM overall, except for the question in the comment

src/page-templates/content-page/content-page-template.tsx
jackyw2018: I do note that I still don't know what are the correct expected values for this field, but perhaps going with the documentation is the safest route for now.

jackyw2018 (approved):

PR #545, opened 2023-07-11
Use RegEx to adapt ContentStack Image Markdown Syntax

jackyw2018:
## Jira Ticket:

[DEVHUB-1821](https://jira.mongodb.org/browse/DEVHUB-1821)

## Updates

- Uses RegEx to adapt the Image markdown syntax `![image description][image url]` into the general markdown syntax `image description`
- Uses RegEx to adapt the Link markdown syntax `[link description][link url]` into the general markdown syntax `[link description](link url)`


mcolella14 (approved):
Very cool! This is a simpler solution than extending our markdown parser, so LGTM.

src/utils/contentstack.ts
```diff
@@ -85,3 +85,62 @@ const isEmptySEO = (seo: { [key: string]: any }) => {
 
     return true;
 };
+
+export const convertCSMarkdownToGeneralMarkdown = (cs_markdown: string) => {
```
harikakotipalli, line 89: @jackyw2018 : Can I get a test case for this one. How are you identifying the citations (meaning numbered references)

jackyw2018: Do you mean to write a test or provide a test case?

I have manually tested by creating content on Contentstack UI and rendering it on DevCenter UI. 

Here is a [CS entry](https://app.contentstack.com/#!/stack/blt39790b633ee0d5a7/content-type/articles/en-us/entry/blte3314a3801e3f4a8/edit?branch=main) which I have:
1. CS-inserted Images in the middle and at the end
2. CS-inserted Link at the end

This appears to render correctly via the preview page.

PR #552, opened 2023-07-20
Update markdown parser

harikakotipalli:
Uses remark-inline-links to handle the 'linkReference' , 'imageReference' and 'definition' node types. Ignores "definition" nodes, converts linkReference to links and imageReference to image.


jackyw2018 (approved):
LGTM!

PR #554, opened 2023-07-27
Playwright tests in drone

mcolella14:
Testing Drone stuff


src/pages/books.tsx
```diff
@@ -93,10 +94,14 @@ const RelatedBookCard: React.FunctionComponent<Book> = ({
                     height: '256px',
                 }}
             >
-                <img
+                <Image
```
harikakotipalli, line 97: is this tag change related to playwright tests ?

.drone.yml
```diff
@@ -1,3 +1,226 @@
...
+    - npm install --global yarn
+    - yarn install --frozen-lockfile
+    - yarn build
+    - export CHROME_PATH=$(which google-chrome-stable)
+    - export LHCI_BUILD_CONTEXT__CURRENT_BRANCH={$DRONE_BRANCH}
+    - yarn global add @lhci/cli@0.12.x
+    - lhci autorun
+    resources:
+      limits:
+        memory: 18Gi
```
harikakotipalli, line 120: Will this memory limit connect to devrel name space ?

harikakotipalli: if yes, try to give optimal value. Kanopy team is going to send resource utilization converted into $$$ equivalent and send to org owners.

harikakotipalli (approved):

.drone.yml
mcolella14: @harikakotipalli Ahh ok I can do that. Do you have any estimate of what would be a good limit here?

harikakotipalli: here are the efficiency dashboards for [staging](https://grafana.staging.corp.mongodb.com/d/43f3f1511e5d14ecd08f95e773357949/namespace-resource-efficiency?orgId=1&var-datasource=default&var-namespace=devrel&var-num_pods=10&var-workload=All&var-instance=All&var-component=All&var-version=All) and [prod](https://grafana.corp.mongodb.com/d/43f3f1511e5d14ecd08f95e773357949/namespace-resource-efficiency?orgId=1&var-datasource=default&var-namespace=devrel&var-num_pods=40&var-workload=All&var-instance=All&var-component=All&var-version=All)

mcolella14: Actually since these are being run in drone containers I don't think they will be in the devrel namespace. I will adjust though, 18G seems like a lot.

PR #570, opened 2023-08-07
Fix jest tests

mcolella14:
## Description:

Removed visibility checks on the FollowLink component since there is something wrong with the `toBeVisible` method.

Also updated the mocks that were giving us issues.


harikakotipalli (approved):
LGTM

jackyw2018 (approved):

PR #590, opened 2023-08-18
[FIX] Account for content with no image in live previews

mcolella14:
## Description:
Account for content with no image in live previews


harikakotipalli (approved):

jackyw2018 (approved):

src/service/build-content-items.ts
```diff
@@ -327,15 +327,11 @@ export const CS_previewMapPreviewArticleToContentItem = (
         githubUrl: github_url,
         liveSiteUrl: livesite_url,
         seo: mapSEO(a.seo) as SEO,
-        image: {
-            url: a.image.url,
-            alt: a.image.description || '',
-        },
     };
 
     if (a.image) {
```
harikakotipalli, line 332: what is a.image is returning even ?

PR #605, opened 2023-09-06
DEVHUB-1892 dropping the last bread crumb from past events bread crum…

harikakotipalli:
https://jira.mongodb.org/browse/DEVHUB-1892

### DIFF
--- a/src/page-templates/content-page/content-page-template.tsx
+++ b/src/page-templates/content-page/content-page-template.tsx
@@ -620,11 +620,12 @@ const ContentPageTemplate: NextPage<ContentPageProps> = ({
         );
     };
 
-    const isPastEvent = (contentDate: string | [string, string]) => {
+    const isPastEvent = (date: string | [string, string]) => {
         return !(
-            contentDate.length == 2 &&
-            contentDate[1] &&
-            new Date(contentDate[1]) > new Date()
+            Array.isArray(date) &&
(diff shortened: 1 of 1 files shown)


src/page-templates/content-page/content-page-template.tsx
```diff
@@ -620,6 +620,14 @@ const ContentPageTemplate: NextPage<ContentPageProps> = ({
...
     };
 
+    const isPastEvent = (contentDate: string | [string, string]) => {
+        return !(
+            contentDate.length == 2 &&
+            contentDate[1] &&
+            new Date(contentDate[1]) > new Date()
+        );
+    };
+
```
jackyw2018, line 630: Perhaps we should check if `contentDate` is an array to avoid the unexpected behavior when `contentDate` is a string. Also, should we handle the case when contentDate is a string?

```suggestion
    const isPastEvent = (contentDate: string | [string, string]) => {
        const endDate = Array.isArray(contentDate) ? contentDate[1] : contentDate
        return  new Date(endDate) < new Date()
    };

```

harikakotipalli: For industry events its always guaranteed to be array as end date is a required field in CMS , but good catch and I should handle this , since this will lead to page crashing if string type gets passed for whatever reason.

mcolella14 (approved):

PR #612, opened 2023-09-20
DEVHUB-1186 setting the codeblock height to auto for all codeblocks, …

harikakotipalli:
https://jira.mongodb.org/browse/DEVHUB-1186 - this is a old ticket. This might not be making to production - but it has to be in production in order to share with devrel folks and get their feedback. Will rollback if they don't want this.

### DIFF
--- a/src/components/article-body/body-components/codepanel.tsx
+++ b/src/components/article-body/body-components/codepanel.tsx
@@ -1,9 +1,5 @@
 import { CodePanel } from '@mdb/flora';
 import { getFloraLanguage } from '../../../utils/get-flora-language';
-import {
-    getCodeHeight,
-    getNumberOfLines,
-} from '../../../utils/get-code-height';
 import { LanguageModeType } from '@mdb/flora/dist/types/src/CodeSnippet/types';
 
 type Snippet = {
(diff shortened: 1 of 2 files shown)


jackyw2018 (approved):

src/components/article-body/body-components/codepanel.tsx
```diff
@@ -1,52 +1,92 @@
...
+                        background: '#e7eeec',
+                    },
+                }}
+                ref={wrapper}
+            >
+                <CodePanel
+                    snippets={snippets}
+                    height="auto"
+                    lineWrapping={false}
+                />
```
harikakotipalli, line 88: @mcolella14 for some reason the rendering of code tabs is not matching the story book of flora - not sure why - can you please check. 
image
image

src/components/article-body/body-components/codepanel.tsx
```diff
@@ -1,52 +1,92 @@
...
-                lineWrapping={false}
-            />
+            <div
+                sx={{
+                    '.CodeMirror-simplescroll': { isolation: 'isolate' },
+                    '.CodeMirror-simplescroll-horizontal div': {
+                        background: '#e7eeec',
+                    },
+                }}
+                ref={wrapper}
```
harikakotipalli, line 82: @mcolella14 : I have taken this workaround for codemirror issue from codeblock.tsx , I'm assuming the fix isolation: isolate should be applied to code panel as well.

PR #620, opened 2023-09-26
Update sentry org for migration purposes

jackyw2018:
This merge will only take place after @ajhenri leaves a comment in this PR to signal the Sentry transfer.


ajhenri:
@jackyw2018 You are good to go on this now. You'll also need to update the DSN to the one in https://mongodb-org.sentry.io/settings/projects/devcenter/keys/.

ajhenri (approved):

PR #645, opened 2023-10-30
DEVHUB-1938 pulling original publish date from publish metadata colle…

harikakotipalli:
pulling original publish date from publish metadata collection
https://jira.mongodb.org/browse/DEVHUB-1938

Backend PR : https://github.com/10gen/devcenter-backend/pull/127

A new collection "Publish Metadata" is created in contentstack which will store the publish date of the article when published from "ContentStack". There is a realm webhook to create metadata entry upon publishing  which needs to be migrated later as captured in https://jira.mongodb.org/browse/DEVHUB-1938

### DIFF
--- a/src/api-requests/get-articles.ts
+++ b/src/api-requests/get-articles.ts
@@ -1,6 +1,10 @@
 import { getAllArticlesQuery, getArticleQuery } from '../graphql/articles';
 import { CS_ArticleResponse } from '../interfaces/article';
 import { fetchAll, getClient } from './contentstack_utils';
+import { getPublishingMetadataQuery } from '../graphql/publish-metdata';
+import { CS_PublishMetadataResponse } from '../interfaces/publish-metadata';
+import { ApolloClient, NormalizedCacheObject } from '@apollo/client';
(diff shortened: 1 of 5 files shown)


src/api-requests/get-articles.ts
```diff
@@ -26,7 +70,7 @@ export const CS_getArticleBySlugFromCMS = async (
         client,
         variables
     )) as CS_ArticleResponse[];
-
+    await CS_adjustOriginalPublishDate(client, calculatedSlug, articles);
```
mcolella14, line 73: Do we have a sense of how this will affect page speed? Seems like fetching both all articles and all publish metadata might add some time.

mcolella14 (approved):
LGTM, but just curious about latency.

src/api-requests/get-articles.ts
harikakotipalli: it will add some time - let me check the metrics - 
there are 2 other options 
1. get from preval - which might be stale
2. combine the graphql query - like a join

